PR #7, opened 2020-07-29
Magic wand floodfill

charmychancey:
floodfill documentation at a glance:

Set<number> floodfill(imgData: ImageData, x: number, y: number, tolerance: number)

Input a coordinate (x, y values) and tolerance level to create a mask based off of the original image given by imgData.
The mask is represented as a set of numbers that represent the indices of pixels belonging to that mask.

### DIFF
--- a/src/main/ngapp/accio/src/app/editor/magic-wand.service.spec.ts
+++ b/src/main/ngapp/accio/src/app/editor/magic-wand.service.spec.ts
@@ -13,4 +13,121 @@ describe('MagicWandService', () => {
   it('should be created', () => {
     expect(service).toBeTruthy();
   });
+  it('Test method: coordToDataArrayIndex()', () => {
+    // Should convert [x][y] to 1-D indexing
+    expect(service.coordToDataArrayIndex(
+        0, 0, 100)).toEqual(0);
+    expect(service.coordToDataArrayIndex(
+        1, 5, 100)).toEqual(2004);
+  });
+  it('Test method: dataArrayToRgba()', () => {
+    // Should access RGBA values of img given a coord
+    const imgData: ImageData = new ImageData(11, 4);
+    imgData.data[56] = 120;
+    imgData.data[56 + 1] = 111;
+    imgData.data[56 + 2] = 117;
+    imgData.data[56 + 3] = 255;
+    expect(service.dataArrayToRgba(
+        imgData, 3, 1)).toEqual([120, 111, 117, 255]);
+  });
+  it('Test method: isInBounds() >> top left OutOfRange', () => {
+    // Should return false if inputted pixel's coord is out of range of the img
+    const imgData: ImageData = new ImageData(11, 4);
+    expect(service.isInBounds(
+        imgData.width, imgData.height, -1, 1)).toEqual(false);
+  });
+  it('Test method: isInBounds() >> top right OutOfRange', () => {
+    const imgData: ImageData = new ImageData(11, 4);
+    expect(service.isInBounds(
+      imgData.width, imgData.height, 11, 1)).toEqual(false);
+  });
+  it('Test method: isInBounds() >> bot left OutOfRange', () => {
+    const imgData: ImageData = new ImageData(11, 4);
+    expect(service.isInBounds(
+      imgData.width, imgData.height, 4, -1)).toEqual(false);
+  });
+  it('Test method: isInBounds() >> bot right OutOfRange', () => {
+    const imgData: ImageData = new ImageData(11, 4);
+    expect(service.isInBounds(
+      imgData.width, imgData.height, 4, 4)).toEqual(false);
+  });
+  it('Test method: isInBounds() >> in range', () => {
+    // Should return true if inputted pixel's coords in in range of img
+    const imgData: ImageData = new ImageData(11, 4);
+    expect(service.isInBounds(
+      imgData.width, imgData.height, 4, 2)).toEqual(true);
+  });
+  it('Test method: getIsValid() >> not valid', () => {
+    // Should return false if isInBounds() but already visited
+    const visited: Set<number> = new Set([20]);
+    expect(service.getIsValid(11, 4, 5, 0, visited)).toEqual(false);
+  });
+  it('Test method: getIsValid() >> valid', () => {
+    // Should return true if isInBounds() and not yet visited
+    const visited: Set<number> = new Set([24]);
+    expect(service.getIsValid(11, 4, 5, 0, visited)).toEqual(true);
+  });
+  it('Test method: getIsMask() >> not valid', () => {
+    // Should return false if pixel of interest is outside of tolerance range
+    const visited: Set<number> = new Set();
+    const imgData: ImageData = new ImageData(11, 4);
+    imgData.data[24] = 120;
+    imgData.data[24 + 1] = 117;
+    imgData.data[24 + 2] = 108;
+    imgData.data[24 + 3] = 255;
+    expect(service.getIsMask(
+        [131, 117, 108, 255], imgData, [6, 0], 10, visited)).toEqual(false);
+  });
+  it('Test method: getIsMask() >> valid', () => {
+    // Should return true if pixel of interest is within tolerance range
+    const visited: Set<number> = new Set();
+    const imgData: ImageData = new ImageData(11, 4);
+    imgData.data[24] = 120;
+    imgData.data[24 + 1] = 117;
+    imgData.data[24 + 2] = 108;
+    imgData.data[24 + 3] = 255;
+    expect(service.getIsMask(
+      [130, 117, 108, 255], imgData, [6, 0], 10, visited)).toEqual(true);
+  });
+  it('Test method: floodfill() >> no adjacent pixels in mask', () => {
+    // Should only return index of pixel that was initially clicked
+    const imgData: ImageData = new ImageData(11, 4);
+    const notMask: Set<number> = new Set(
+        [8, 24, 28, 36, 48, 56, 64, 76, 96, 100, 112, 116, 152]);
+    for (const pixel of notMask) {
+      imgData.data[pixel] = 255;
+      imgData.data[pixel + 1] = 255;
+      imgData.data[pixel + 2] = 255;
+      imgData.data[pixel + 3] = 255;
+    }
+    expect(service.floodfill(imgData, 2, 1, 1)).toEqual(new Set<number>([52]));
+  });
+  it('Test method: floodfill() >> small mask, touches border', () => {
+    // Should return a mask covering the bottom right corner of the img
+    const imgData: ImageData = new ImageData(11, 4);
+    const notMask: Set<number> = new Set(
+        [8, 24, 28, 36, 48, 56, 64, 76, 96, 100, 112, 116, 152]);
+    for (const pixel of notMask) {
+      imgData.data[pixel] = 255;
+      imgData.data[pixel + 1] = 255;
+      imgData.data[pixel + 2] = 255;
+      imgData.data[pixel + 3] = 255;
+    }
+    expect(service.floodfill(imgData, 8, 2, 1)).toEqual(new Set<number>(
+        [40, 80, 84, 120, 124, 128, 156, 160, 164, 168, 172]));
+  });
+  it('Test method: floodfill() >> small mask in middle of img', () => {
+    // Should return a mask covering the bottom right corner of the img
+    const imgData: ImageData = new ImageData(11, 4);
+    const notMask: Set<number> = new Set(
+      [8, 24, 28, 36, 48, 56, 64, 76, 96, 100, 112, 116, 152]);
+    for (const pixel of notMask) {
+      imgData.data[pixel] = 255;
+      imgData.data[pixel + 1] = 255;
+      imgData.data[pixel + 2] = 255;
+      imgData.data[pixel + 3] = 255;
+    }
+    expect(service.floodfill(imgData, 6, 1, 1)).toEqual(new Set<number>(
+      [68, 72]));
+  });
 });
--- a/src/main/ngapp/accio/src/app/editor/magic-wand.service.spec.ts
+++ b/src/main/ngapp/accio/src/app/editor/magic-wand.service.spec.ts
@@ -68,7 +68,7 @@ describe('MagicWandService', () => {
     const imgData: ImageData = makeTestImage(
         11, 4, [120, 117, 108, 255], new Set<number>([24]));
     expect(service.getIsMask(
-        [131, 117, 108, 255], imgData, [6, 0], 10, new Set<number>())).toEqual(
+        [131, 117, 108, 255], imgData, [6, 0], 10)).toEqual(
         false);
   });
   it('Test method: getIsMask() >> valid', () => {
@@ -76,7 +76,7 @@ describe('MagicWandService', () => {
     const imgData: ImageData = makeTestImage(
         11, 4, [120, 117, 108, 255], new Set<number>([24]));
     expect(service.getIsMask(
-        [130, 117, 108, 255], imgData, [6, 0], 10, new Set<number>())).toEqual(
+        [130, 117, 108, 255], imgData, [6, 0], 10)).toEqual(
         true);
   });
   it('Test method: floodfill() >> no adjacent pixels in mask', () => {
--- a/src/main/ngapp/accio/src/app/editor/magic-wand.service.spec.ts
+++ b/src/main/ngapp/accio/src/app/editor/magic-wand.service.spec.ts
@@ -1,0 +1,16 @@
+import { TestBed } from '@angular/core/testing';
+
+import { MagicWandService } from './magic-wand.service';
+
+describe('MagicWandService', () => {
+  let service: MagicWandService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({});
+    service = TestBed.inject(MagicWandService);
+  });
+
+  it('should be created', () => {
+    expect(service).toBeTruthy();
+  });
+});
--- a/src/main/ngapp/accio/src/app/editor/magic-wand.service.spec.ts
+++ b/src/main/ngapp/accio/src/app/editor/magic-wand.service.spec.ts
@@ -2,6 +2,8 @@ import { TestBed } from '@angular/core/testing';
 
 import { MagicWandService } from './magic-wand.service';
 
+
+/* tslint:disable */
 describe('MagicWandService', () => {
   let service: MagicWandService;
 
@@ -23,44 +25,38 @@ describe('MagicWandService', () => {
   it('Test method: dataArrayToRgba()', () => {
     // Should access RGBA values of img given a coord
     const imgData: ImageData = new ImageData(11, 4);
-    imgData.data[56] = 120;
-    imgData.data[56 + 1] = 111;
-    imgData.data[56 + 2] = 117;
-    imgData.data[56 + 3] = 255;
+    imgData.data[56] = 120;  // Red
+    imgData.data[56 + 1] = 111;  // Green
+    imgData.data[56 + 2] = 117;  // Blue
+    imgData.data[56 + 3] = 255;  // Alpha
     expect(service.dataArrayToRgba(
         imgData, 3, 1)).toEqual([120, 111, 117, 255]);
   });
   it('Test method: isInBounds() >> top left OutOfRange', () => {
     // Should return false if inputted pixel's coord is out of range of the img
-    const imgData: ImageData = new ImageData(11, 4);
     expect(service.isInBounds(
-        imgData.width, imgData.height, -1, 1)).toEqual(false);
+        11, 4, -1, 1)).toEqual(false);
   });
   it('Test method: isInBounds() >> top right OutOfRange', () => {
-    const imgData: ImageData = new ImageData(11, 4);
     expect(service.isInBounds(
-      imgData.width, imgData.height, 11, 1)).toEqual(false);
+        11, 4, 11, 1)).toEqual(false);
   });
   it('Test method: isInBounds() >> bot left OutOfRange', () => {
-    const imgData: ImageData = new ImageData(11, 4);
     expect(service.isInBounds(
-      imgData.width, imgData.height, 4, -1)).toEqual(false);
+        11, 4, 4, -1)).toEqual(false);
   });
   it('Test method: isInBounds() >> bot right OutOfRange', () => {
-    const imgData: ImageData = new ImageData(11, 4);
     expect(service.isInBounds(
-      imgData.width, imgData.height, 4, 4)).toEqual(false);
+        11, 4, 4, 4)).toEqual(false);
   });
   it('Test method: isInBounds() >> in range', () => {
     // Should return true if inputted pixel's coords in in range of img
-    const imgData: ImageData = new ImageData(11, 4);
     expect(service.isInBounds(
-      imgData.width, imgData.height, 4, 2)).toEqual(true);
+        11, 4, 4, 2)).toEqual(true);
   });
   it('Test method: getIsValid() >> not valid', () => {
     // Should return false if isInBounds() but already visited
-    const visited: Set<number> = new Set([20]);
-    expect(service.getIsValid(11, 4, 5, 0, visited)).toEqual(false);
+    expect(service.getIsValid(11, 4, 5, 0, new Set([20]))).toEqual(false);
   });
   it('Test method: getIsValid() >> valid', () => {
     // Should return true if isInBounds() and not yet visited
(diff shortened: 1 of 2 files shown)


src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,159 @@
...
+      if (failsTolerance) {
+        return !isStillMask;
+      }
+    }
+
+    return isStillMask;
+  }
+
+  // Check if @pixelCoord is in bounds and makes sure it's not a repeat coord
+  getIsValid(imgData: ImageData, pixelCoord: Array<number>, visited: Set<number>): boolean {
```
dtjanaka, line 110: Seems expensive to pass the entire ImageData object for every call of `getIsValid`. Looks like only the dimensions are needed for this function. Similarly for other functions, try to minimize the amount of information passed around.

charmychancey: Change request addressed in the commits below.

On a side note: can we also discuss the cost of of switching from imgData >> imgData.width, imgData.height as args for the method for the purposes of future code design? 
First, my assumption was that having less arguments as parameters for methods would maintain cleaner code as far as readability, as well as inputting broader data types as arguments would keep the code more flexible if it needs to be altered down the line.
Second, I was wondering if the difference between passing in ImageData as a param as opposed to passing in imaData.width and imgData.height as params are actually negligible in terms of operational expense. 
For example, by passing in imgData: ImageData as a parameter, I thought we are just passing in a single address in memory that holds the value of an address as a variable reference to the actual ImageData object, and thus is approx. the same as passing around imgData.width and imgData.height, which are two separate addresses in memory that hold similar things respectively. Based on this assumption, how I understand it is that the difference is equivalent to accessing 1 extra memory address in the code per use of getIsValid(), which leads to O(n + 1) is O(n). 
I'm also mainly asking because this might be a good opportunity for me to learn from you, because my impression is that you might be more familiar with C than me, which gives me the assumption that you might also understand the little nuances of costs of operations more than me.

charmychancey:
@shmcaffrey This is the .ts code I had to write in order to setup a testing environment for floodfill(). It doesn't belong in the pull request code itself, but you might still find it useful if you want to repurpose some of it for the UI side

// Return [x,y] mouse coordinate within the @canvas
function getMousePosition(canvas: HTMLCanvasElement, event: MouseEvent): Array<number> {
  let rect: ClientRect = canvas.getBoundingClientRect();
  let x: number = event.clientX - rect.left;
  let y: number = event.clientY - rect.top;
 
  return [x, y];
}
 
// Write @msg inside the @canvas at top left corner
function writeMsg(canvas: HTMLCanvasElement, msg) {
  let context = canvas.getContext('2d');
  context.clearRect(0, 0, 250, 40);
  // context.font = '18pt Calibri';
  context.fillStyle = 'black';
  context.fillText(msg, 10, 10);
}
 
// Convert coord [@x, @y] to indexing style of imageData.data
function coordToDataArrayIndicies(x: number, y: number, width: number): Array<number> {
  // Equivalent to: (x * 4) + (y * width * 4)
  let red: number = (x + (y * width)) * 4;
  return [red, red + 1, red + 2, red + 3];
}
 
// This is the function called by index.html
function drawMask() {
  let canvas = document.getElementById('canvas') as HTMLCanvasElement;
  let context = canvas.getContext('2d');
  let defaultImg: HTMLImageElement = new Image();
  defaultImg.src = 'images/iu-wallpaper-1.jpg';
  // Draw the default image in the canvas first
  defaultImg.onload = function() {
    context.drawImage(defaultImg, 0, 0);
  };
 
  // Get coordinate where mouse is over, dislay on top left of canvas
  canvas.addEventListener('mousemove', function(event) {
    let mousePos: Array<number> = getMousePosition(canvas, event);
    let msg: string = `Mouse position: ${mousePos[0]}, ${mousePos[1]}`;
    writeMsg(canvas, msg);
  }, false);
 
  // Draw the mask
  canvas.addEventListener('click', function(event) {
    let mousePos: Array<number> = getMousePosition(canvas, event);
 
    let myImgData: ImageData = context.getImageData(0, 0, canvas.width, canvas.height);
    let mask: Set<number> = floodfill(myImgData, mousePos[0], mousePos[1], 20);
    console.log(mask);
 
    for (let pixelIndex of mask) {
      //TODO
      // let x: number = pixelCoord[0];
      // let y: number = pixelCoord[1];
      // let pixelIndex: Array<number> = coordToDataArrayIndicies(x, y, canvas.width);
 
      let redIndex: number = pixelIndex;
      let greenIndex: number = pixelIndex + 1;
      let blueIndex: number = pixelIndex + 2;
      let alphaIndex: number = pixelIndex + 3;
 
      myImgData.data[redIndex] = 128;
      myImgData.data[greenIndex] = 128;
      myImgData.data[blueIndex] = 128;
      myImgData.data[alphaIndex] = 255;
      // console.log(myImgData[redIndex]);
    }
    console.log('Putting new image...');
    context.putImageData(myImgData, 0, 0);
    console.log('Finished putting new image...');
  });
}

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,159 @@
...
+    let red: number = data[pixelIndex];
+    let green: number = data[pixelIndex + 1];
+    let blue: number = data[pixelIndex + 2];
+    let alpha: number = data[pixelIndex + 3];
+    // Store original pixel's attributes
+    let pixel: Array<number> = [red, green, blue, alpha];
+    return pixel;
+  }
+
+  // Convert coord [@x, @y] to indexing style of imageData.data
```
shmcaffrey, line 148: might add in the comments that this assumes x & y are coordinates in the imageData

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,159 @@
...
+    // Use a set for mask; mainly do iter and set operations on masks
+    const mask: Set<number> = new Set();
+    // Represent [R,G,B,A] attributes of initial pixel
+    const originalPixel: Array<number> = this.dataArrayToRGBA(imgData, xCoord, yCoord);
+
+    visit.push([xCoord, yCoord]);
+    // Convert [x,y] format coord to 1-D equivalent of imgData.data (DataArray)
+    let indexAsDataArray: number = this.coordToDataArrayIndicies(xCoord, yCoord, imgData.width)[0];
+    visited.add(indexAsDataArray);
+    let visitSize: number = 1;
```
shmcaffrey, line 26: I think sets have a size property? visited.size, is there a reason to store the size separate?

charmychancey: I was thinking about the future. When coming back to optimize after we finish the mvp, an idea I had was to declare the array with a bigger default initialized size. If we decide to do this in the future, the .size property wouldn't work/accurately reflect what we want it to anymore.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
charmychancey: I don't think that necessarily applies here. This is only a general helper function to convert from a 2-D array to a 1-D array indexing style, and so it's purpose is not explicitly to interact with an ImageData object on it's own.

charmychancey: Instead, I will change comment to not suggest it is connected to ImageData but instead  DataArray

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
dtjanaka: I was unclear about when JS passes by reference vs value. LGTM

robotsaurus: In JS objects and arrays are passed by reference and otherwise value.  By being implicit it leads to this uncertainty.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    let red: number = data[pixelIndex];
+    let green: number = data[pixelIndex + 1];
+    let blue: number = data[pixelIndex + 2];
+    let alpha: number = data[pixelIndex + 3];
+    // Store original pixel's attributes
+    let pixel: Array<number> = [red, green, blue, alpha];
+    return pixel;
+  }
+
+  // Convert coord [@x, @y] (2-D) to indexing style of DataArray (1-D)
```
robotsaurus, line 148: From the comment I didn't realize you were returning the RGBA indices.  I thought it as just the start of the pixel so it would be good to clarify that.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+  /** Return set of coordinates(formatted as a 1-D array).
+   *  Coordinates correspond to pixels considered part of the mask.
+   */
+  /* tslint:disable */
+  floodfill(imgData: ImageData, xCoord: number, yCoord: number,
+            tolerance: number): Set<number> {
+    // Use basic queue to immitate recursive approach of the stack
+    // possible TODO: set array size after mvp to optimize heap memory access
+    const visit: Array<Array<number>> = new Array();
+    const visited: Set<number> = new Set();
```
robotsaurus, line 16: It's a little inconsistent that you store what pixels to visit as the 2D x,y location but visited is the index of the R channel in the 1D array.  Is there a reason for this choice?

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    // Use a set for mask; mainly do iter and set operations on masks
+    const mask: Set<number> = new Set();
```
robotsaurus, line 18: I assume this stores the index of the R pixel?

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
robotsaurus: I would say that this is "future proofing code".  By doing this you are now requiring that whenever someone adds/removes something from the array they need to remember to increment/decrement the visitSize.   It is better to use the .size method because then it will never become out of sync.  If/when you implement your performance optimization then you would be trading off the complexity for performance.  Now you are just increasing the complexity for no gain.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+
+      // Loop part of while-loop
+
+      // Get coords of adjacent pixels
+      let leftPixelCoord: Array<number> = [x - 1, y];
+      let rightPixelCoord: Array<number> = [x + 1, y];
+      let upPixelCoord: Array<number> = [x, y - 1];
+      let downPixelCoord: Array<number> = [x, y + 1];
+      // Add coords of adjacent pixels to the heap
+      let isStillMask: boolean = this.getIsStillMask(originalPixel, imgData, leftPixelCoord, tolerance, visited);
```
robotsaurus, line 47: 'Still' in your variable and function names is a bit awkward.  I think removing it makes it more readable and clear.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+        visitSize += 1;
+      }
+    }  // end of while loop
+
+    return mask;
+  }
+
+  /** Judge current pixel's RGBA against original pixel's RGBA to see if it
+   *  can still be part of the mask (using tolerance criteria)
+   */
```
robotsaurus, line 78: Add that you additionally check whether the pixel is in the image

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+  getIsStillMask(originalPixel: Array<number>, imgData: ImageData,
+                         pixelCoord: Array<number>, tolerance: number,
+                         visited: Set<number>): boolean {
```
robotsaurus, line 81: nit: weird indentation

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+    // Do not push repeat coords to heap
+    let indexAsDataArray: number = this.coordToDataArrayIndices(curX, curY, imgWidth)[0];
+    if (visited.has(indexAsDataArray)) {
+      return !isValid;
+    }
+    visited.add(indexAsDataArray);
+
+    return isValid;
+  }
```
robotsaurus, line 129: I think it's better to split this into two functions:

1. Check whether it is a valid image coordinate.
2. Check whether it has been visited before.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    let xOutOfBounds: boolean = curX < 0 || curX > imgWidth - 1;
+    if (yOutOfBounds || xOutOfBounds) {
+      return !isValid;
+    }
+    // Do not push repeat coords to heap
+    let indexAsDataArray: number = this.coordToDataArrayIndices(curX, curY, imgWidth)[0];
+    if (visited.has(indexAsDataArray)) {
+      return !isValid;
+    }
+    visited.add(indexAsDataArray);
```
robotsaurus, line 126: I didn't expect this function to also add it to the visited array. I would move that out.

You want to make sure functions have a clear and distinct purpose.  It seems like the purpose of this function checks whether the pixel is valid so adding it to the visited array seems like a tangential side affect.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+  getIsStillMask(originalPixel: Array<number>, imgData: ImageData,
+                         pixelCoord: Array<number>, tolerance: number,
+                         visited: Set<number>): boolean {
+    let isStillMask: boolean = true;
+
+    // Preface; check if pixel is valid (indexing errs and repeat values)
+    let isValid: boolean = this.getIsValid(imgData.width, imgData.height, pixelCoord, visited);
+    if (!isValid) {
+      // Automatically not in mask b/c failed vailidity test
+      return !isStillMask;
```
robotsaurus, line 88: more clear to just 

`return false;`

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    let curY: number = pixelCoord[1];
+    let curPixel: Array<number> = this.dataArrayToRGBA(imgData, curX, curY);
+
+    // All attributes of the pixel (R,G,B, and A) must be within tolerance level
+    for (let i = 0; i < 4; i++) {
+      let upperTolerance: boolean = curPixel[i] > originalPixel[i] + tolerance;
+      let lowerTolerance: boolean = curPixel[i] < originalPixel[i] - tolerance;
+      let failsTolerance: boolean = upperTolerance || lowerTolerance;
+      if (failsTolerance) {
+        return !isStillMask;
```
robotsaurus, line 102: more clear to just

`return false;`

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+      }
+    }
+
+    return isStillMask;
```
robotsaurus, line 106: isStillMask is never changed from true so here it is better to just return true;

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    // Get array of attributes of current pixel
+    let curX: number = pixelCoord[0];
+    let curY: number = pixelCoord[1];
+    let curPixel: Array<number> = this.dataArrayToRGBA(imgData, curX, curY);
+
+    // All attributes of the pixel (R,G,B, and A) must be within tolerance level
+    for (let i = 0; i < 4; i++) {
+      let upperTolerance: boolean = curPixel[i] > originalPixel[i] + tolerance;
+      let lowerTolerance: boolean = curPixel[i] < originalPixel[i] - tolerance;
+      let failsTolerance: boolean = upperTolerance || lowerTolerance;
```
robotsaurus, line 100: no need to introduce an additional variable for this.  It requires the code reader to have to remember what condition this satisfies when inside the if() statement.  Since it isn't a complicated expression I would just put it directly in.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    let indexAsDataArray: number = this.coordToDataArrayIndices(curX, curY, imgWidth)[0];
+    if (visited.has(indexAsDataArray)) {
+      return !isValid;
+    }
+    visited.add(indexAsDataArray);
+
+    return isValid;
+  }
+
+  // Return pixel attributes of @imgData at [@xCoord, @yCoord] as Array<number>
```
robotsaurus, line 131: specify that you return RGBA in that order.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+  dataArrayToRGBA(imgData: ImageData, xCoord: number, yCoord: number): Array<number> {
+    // Unpack imgData for readability
+    let data: Uint8ClampedArray = imgData.data;
```
robotsaurus, line 134: is this a reference otherwise it's a costly copy.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    let imgWidth: number = imgData.width;
+
+    // Pixel attributes in imgData are organized adjacently in a 1-D array
+    let pixelIndex: number = (xCoord + (imgWidth * yCoord)) * 4;
```
robotsaurus, line 138: you should use your utility function `coordToDataArrayIndices ` here

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class MagicWandService {
+  /** Return set of coordinates(formatted as a 1-D array).
+   *  Coordinates correspond to pixels considered part of the mask.
+   */
+  /* tslint:disable */
```
robotsaurus, line 10: You should clarify that this does a 4-way floodfill as opposed to an 8-way.

robotsaurus:
I don't see any unit tests for these functions.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
charmychancey: I tried to leave it in x,y format as much as possible and convert between the two formats as little as possible in order to make passing the parameters more consistent and the code easier to look at in general. 

However, I had to omit that for the visited variable. If I had passed in an [x,y] into visited, it would have been a Set<Array<number>> instead of a Set<number> which I had trouble with because doing set operations made it compare array objects instead of the actual coordinates. I then thought it would be simplest to pass the index of R channel into visited instead of passing pairs of numbers into visited.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
charmychancey: visited stores the index of the start of the pixel representation in the DataArray, and yes the start of the pixel representation is the attribute for red of that pixel.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
charmychancey: I'll go ahead remove the complexity for now then.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
charmychancey: I believe it is a pass by reference

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
charmychancey: Ok, I went ahead and broke the logic up using two helper functions

charmychancey:
ptal

charmychancey:
> I don't see any unit tests for these functions.
Just want to check the code in first to unblock @shmcaffrey. Will follow up with units tests asap

charmychancey:
Ptal

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,169 @@
...
+    const mask: Set<number> = new Set();
+    // Represent [R,G,B,A] attributes of initial pixel
+    const originalPixel: Array<number> = this.dataArrayToRGBA(imgData, xCoord, yCoord);
+
+    visit.push([xCoord, yCoord]);
+    // Convert [x,y] format coord to 1-D equivalent of imgData.data (DataArray)
+    let indexAsDataArray: number = this.coordToDataArrayIndices(xCoord, yCoord, imgData.width)[0];
+    visited.add(indexAsDataArray);
+
+    // Loop until no more adjacent pixel's within tolerance level
```
snazzorama, line 28: pixel's -> pixels

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  // Check bounds of indexing for img dimensions
+  isInBounds(imgWidth: number, imgHeight: number, curX: number, curY: number): boolean {
+    let yOutOfBounds: boolean = curY < 0 || curY > imgHeight - 1;
+    let xOutOfBounds: boolean = curX < 0 || curX > imgWidth - 1;
+    if (yOutOfBounds || xOutOfBounds) {
+      return false;
+    }
+
+    return true;
```
snazzorama, line 125: If you have an "if" block that just returns true or false, it's generally better to just return the value directly:

return !(yOutOfBounds || xOutOfBounds);

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,169 @@
...
+    if (visited.has(index)) {
+      return false;
+    }
+
+    return true;
+  }
+
+
+  // Return pixel attributes of @imgData at [@xCoord, @yCoord] as [R, G, B, A]
+  dataArrayToRGBA(imgData: ImageData, xCoord: number, yCoord: number):
```
snazzorama, line 143: nit: this should be called dataArrayToRgba

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,169 @@
...
+    // Store original pixel's attributes
+    return [red, green, blue, alpha];
+  }
+
+  // Convert coord [@x, @y] (2-D) to indexing style of DataArray (1-D)
+  /**Returns [indexOfR, indexOfG, indexOfB, indexOfA]
+   * Important: Returns array of number representing indices.
+   * First element contains value of index for attribute: red and so on
+   * */
+  coordToDataArrayIndices(x: number, y: number, width: number): Array<number> {
```
snazzorama, line 165: In every caller except one, you are just using the first element of the returned array. I would suggest having this function just return the first index, and then doing the +1, +2, etc in that one call site.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,169 @@
...
+      let leftPixelCoord: Array<number> = [x - 1, y];
+      let rightPixelCoord: Array<number> = [x + 1, y];
+      let upPixelCoord: Array<number> = [x, y - 1];
+      let downPixelCoord: Array<number> = [x, y + 1];
+      // Add coords of adjacent pixels to the heap
+      let isStillMask: boolean = this.getIsMask(originalPixel, imgData, leftPixelCoord, tolerance, visited);
+      // Check left pixel
+      if (isStillMask) {
+        visit.push(leftPixelCoord);
+      }
```
snazzorama, line 50: You want to avoid cut-and-pasting code as much as possible. For example, if you ever want to change one of the four copies of this block, you'll have to change all of them. I would suggest:

let neighbors = [[x-1, y], [x+1, y], [x, y-1], [x, y+1]];
for (let neighbor of neighbors) {
 ...
}

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,169 @@
...
+export class MagicWandService {
+  /**4-Way floodfill (left, right, up, down):
+   * Return set of coordinates(formatted as a 1-D array).
+   * Coordinates correspond to pixels considered part of the mask.
+   */
+  /* tslint:disable */
+  floodfill(imgData: ImageData, xCoord: number, yCoord: number,
+            tolerance: number): Set<number> {
+    // Use basic queue to immitate recursive approach of the stack
+    // possible TODO: set array size after mvp to optimize heap memory access
```
snazzorama, line 15: This comment still leaves some questions in my mind about what these variables mean. I'd suggest spelling it out:

We will store a queue of pixels that we need to visit in \`visit\`.
We will store the already-visited pixels in \`visited\`.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    return true;
+  }
+
+  // Check if @pixelCoord is in bounds and makes sure it's not a repeat coord
+  getIsValid(imgWidth: number, imgHeight: number, curX: number, curY: number,
+    visited: Set<number>): boolean {
+    if (this.isInBounds(imgWidth, imgHeight, curX, curY) &&
+      this.notVisited(imgWidth, curX, curY, visited)) {
```
snazzorama, line 110: nit: line wrapping continuation should be +4 spaces

https://google.github.io/styleguide/jsguide.html#formatting-indent

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,169 @@
...
+  notVisited(imgWidth:number, curX: number, curY: number,
+    visited: Set<number>): boolean {
+    // Do not push repeat coords to heap
+    let index: number =
+      this.coordToDataArrayIndices(curX, curY, imgWidth)[0];
+    if (visited.has(index)) {
+      return false;
+    }
+
+    return true;
```
snazzorama, line 138: Same here:

return !visited.has(index);

charmychancey:
Ptal

charmychancey:
Also adding unit test suite for magic-wand up to floodfill() to this PR with my latest commit. Ptal

src/main/ngapp/accio/src/app/editor/magic-wand.service.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+    imgData.data[56 + 2] = 117;
+    imgData.data[56 + 3] = 255;
+    expect(service.dataArrayToRgba(
+        imgData, 3, 1)).toEqual([120, 111, 117, 255]);
+  });
+  it('Test method: isInBounds() >> top left OutOfRange', () => {
+    // Should return false if inputted pixel's coord is out of range of the img
+    const imgData: ImageData = new ImageData(11, 4);
+    expect(service.isInBounds(
+        imgData.width, imgData.height, -1, 1)).toEqual(false);
```
snazzorama, line 37: No need to make an ImageData here; just pass in 11 and 4 to isInBounds().

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    visit.push([xCoord, yCoord]);
+    // Convert [x,y] format coord to 1-D equivalent of imgData.data (DataArray)
+    let indexAsDataArray: number =
+        this.coordToDataArrayIndex(xCoord, yCoord, imgData.width);
+    visited.add(indexAsDataArray);
+
+    // Loop until no more adjacent pixels within tolerance level
+    while(visit.length != 0) {
+      let coord: Array<number> = visit.pop();
```
snazzorama, line 34: const

(Same for many other variables below.)

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    return true;
+  }
+
+  // Check if @pixelCoord is in bounds and makes sure it's not a repeat coord
+  getIsValid(imgWidth: number, imgHeight: number, curX: number, curY: number,
+      visited: Set<number>): boolean {
+    if (this.isInBounds(imgWidth, imgHeight, curX, curY) &&
+        this.notVisited(imgWidth, curX, curY, visited)) {
```
snazzorama, line 99: `return this.isInBounds(imgWidth, imgHeight, curX, curY) && this.notVisited(imgWidth, curX, curY, visited);`

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  // Convert coord [@x, @y] (2-D) to indexing style of DataArray (1-D)
+  /**Returns [indexOfR, indexOfG, indexOfB, indexOfA]
+   * Important: Returns array of number representing indices.
+   * First element contains value of index for attribute: red and so on
+   * */
+  coordToDataArrayIndex(x: number, y: number, width: number): number {
+    let pixelStart: number = (x + (y * width)) * 4;
+    return pixelStart
```
snazzorama, line 151: No need to make a variable if you're just going to return right away:

return (x + (y * width)) * 4;

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    let red: number = data[pixelIndex];
+    let green: number = data[pixelIndex + 1];
+    let blue: number = data[pixelIndex + 2];
+    let alpha: number = data[pixelIndex + 3];
+    // Store original pixel's attributes
+    return [red, green, blue, alpha];
+  }
+
+  // Convert coord [@x, @y] (2-D) to indexing style of DataArray (1-D)
+  /**Returns [indexOfR, indexOfG, indexOfB, indexOfA]
```
snazzorama, line 145: Update this comment.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+      visited: Set<number>): boolean {
+    let curX: number = pixelCoord[0];
+    let curY: number = pixelCoord[1];
+    let imgWidth: number = imgData.width;
+
+    // Preface; check if pixel is valid (indexing errs and repeat values)
+    let isValid: boolean =
+        this.getIsValid(imgWidth, imgData.height, curX, curY, visited);
+    if (!isValid) {
+      // Automatically not in mask b/c failed vailidity test
```
snazzorama, line 74: validity

src/main/ngapp/accio/src/app/editor/magic-wand.service.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+    // Should only return index of pixel that was initially clicked
+    const imgData: ImageData = new ImageData(11, 4);
+    const notMask: Set<number> = new Set(
+        [8, 24, 28, 36, 48, 56, 64, 76, 96, 100, 112, 116, 152]);
+    for (const pixel of notMask) {
+      imgData.data[pixel] = 255;
+      imgData.data[pixel + 1] = 255;
+      imgData.data[pixel + 2] = 255;
+      imgData.data[pixel + 3] = 255;
+    }
```
snazzorama, line 102: It's really hard to know what pixels you're setting. Also, there's a bunch of duplicated code among the tests. I'd suggest making a local helper function in this file, makeTestImage(). It would take in an array of X, Y coordinates, and return an 11x4 ImageData with the given pixel indices set to 255.

charmychancey:
latest change requests addressed. 

ptal

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    return false;
+  }
+
+  // Check bounds of indexing for img dimensions
+  isInBounds(imgWidth: number, imgHeight: number, curX: number, curY: number):
+      boolean {
+    let yOutOfBounds: boolean = curY < 0 || curY > imgHeight - 1;
+    let xOutOfBounds: boolean = curX < 0 || curX > imgWidth - 1;
+
+    return !(yOutOfBounds || xOutOfBounds);
```
robotsaurus, line 108: nit: it is better to use the "positive" boolean expression if you can.

So you could update you checks to be:

```
let yInBounds: boolean = curY >= 0 && curY < imgHeight;
let xInBounds: boolean = curX >= 0 && curX < imgWidth;

return yInBounds && xInBounds;
```

charmychancey:
ptal

PR #8, opened 2020-07-30
add user image to editor canvas

shmcaffrey:


shmcaffrey:
I'm not sure if you get the request for review when it's already accepted, so I just wanted to mention @robotsaurus @snazzorama ptal!

src/main/ngapp/accio/src/app/top-bar/top-bar.component.ts
```diff
@@ -6,14 +6,24 @@ import { Component, OnInit } from '@angular/core';
...
 
-  ngOnInit(): void {
-  }
+  userSignedIn = false;
+  ngOnInit(): void {}
 
-  loggedIn = false;
-  toggleButton() {
-    this.loggedIn = !this.loggedIn;
+  // this needs to be fixed.
```
robotsaurus, line 14: Update comment to explain what needs to be fixed.

robotsaurus: Also is there a reason you don't want to fix it before submitting PR?

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -1,15 +1,50 @@
...
-  ngOnInit(): void {
-  }
+  constructor(
+    private route: ActivatedRoute,
+  ) { }
+
+  ngOnInit() {
+    //returns url user clicked in gallery component
+    this.route.paramMap.subscribe(params => {
+      this.url = imageUrls[+params.get('imgUrl')];
```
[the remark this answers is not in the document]
snazzorama: Style guide says:

"Code must use Number() to parse numeric values, and must check its return for NaN values explicitly, unless failing to parse is impossible from context."

It explicitly forbids using unary + this way.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -1,15 +1,50 @@
...
 
+     this.ctx = this.canvas.nativeElement.getContext('2d');
+     this.draw();
+    });
+  }  
+  
+  // Draws the image user selects from gallery on Canvas
+  // currently image stretches to fit canvas, will adjust so canvas
```
snazzorama, line 37: Please use full sentences that start with a capital letter and end with a period. I can't tell if "currently" is continuing the previous sentence or not.

src/main/ngapp/accio/src/app/top-bar/top-bar.component.ts
```diff
@@ -6,14 +6,24 @@ import { Component, OnInit } from '@angular/core';
...
+    }
+    console.log('User Signed out');
+    this.userSignedIn = false;
+    return signOut;
   }
-}
\ No newline at end of file
+}
+
+export const signIn = 'Sign In';
```
snazzorama, line 28: The names of constants should be cased like this: SIGN_IN

https://google.github.io/styleguide/jsguide.html#naming-constant-names

src/main/ngapp/accio/src/app/top-bar/top-bar.component.ts
```diff
@@ -6,14 +6,24 @@ import { Component, OnInit } from '@angular/core';
...
+
+  toggleButton(): string {
+    if (!this.userSignedIn) { 
+      console.log('User Signed in');
+      this.userSignedIn = true;
+      return signIn;
+    }
+    console.log('User Signed out');
+    this.userSignedIn = false;
+    return signOut;
```
snazzorama, line 24: Out of curiosity, does this return value get used? I see this function used as a callback, but it's not clear to me what the return value does.

src/main/ngapp/accio/src/app/editor/editor.component.ts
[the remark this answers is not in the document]
shmcaffrey: I will work on this, I was following a tutorial on angular and this was the convention they used but I couldn't find any documentation on it. I will change it.

src/main/ngapp/accio/src/app/top-bar/top-bar.component.ts
shmcaffrey: It's a feature that changes the button to say signed out when the user is signed in and visa-versa. I figured the feature is really only important once the Users API is integrated and I didn't want to end up spending too much time trying to fix it before I got the Gallery/Editor pages set up for the main algorithm. It's something I should've just waited to implement at all.

src/main/ngapp/accio/src/app/top-bar/top-bar.component.ts
shmcaffrey: The return value sets what the button says on it. Currently, the button only displays SIGN_IN because the function isn't working properly. I also updated the comment to say this, but because the button doesn't really matter until the User API is set up I pushed off fixing it for now to focus on integrating the main algorithm.

shmcaffrey:
PTAL

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -1,15 +1,63 @@
...
+    });
+
+    this.ctx = this.canvas.nativeElement.getContext('2d');
+    this.draw();
+  }  
+  
+  //  Checks that route for editor URL is in bounds of array of Urls and is a number
+  //  Redirects to gallery component if not
+  private checkImageUrl(index: number) {
+    if (Number.isNaN(index) || index >= imageUrls.length) {
```
robotsaurus, line 42: A Number can be negative right?  So I think you'd also need to check < 0.

shmcaffrey: I guess I'm not sure how the index would be a negative number. Because index comes from the url, which changes depending on what image the user clicks on. If an image is rendered on the gallery page but not in the array, then I could see the use for checking the out of array length, which will come into play with user loaded images, but I'm not sure about a negative. It's not a big check either so it might be fine to leave it in, I was just curious what your thoughts were.

robotsaurus: Yes the index comes from the url but can't the user manually change the number in the url by hand?

shmcaffrey: ah yes, makes sense!

PR #9, opened 2020-07-31
Add UserInfo class, Login servlet, and /blobs POST.

dtjanaka:


src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
...
+    Filter ownEditFilter = new CompositeFilter(
+        CompositeFilterOperator.AND,
+        Arrays.asList(
+            new FilterPredicate("proj-id", FilterOperator.EQUAL, projId),
+            new CompositeFilter(
+                CompositeFilterOperator.OR,
+                Arrays.<Filter>asList(
+                    new FilterPredicate("owners", FilterOperator.EQUAL, uid),
+                    new FilterPredicate("editors", FilterOperator.EQUAL,
+                                        uid)))));
```
charmychancey, line 95: No change request here, but can you explain to me what this line does?

src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
...
+    String uid = userService.getCurrentUser().getUserId();
+    Key assetParentKey = projKey;
+    Entity imgEntity = new Entity("Image", projKey);
 
-    // Get the name entered by the user
-    String name = request.getParameter("name");
+    // Must be an owner or editor
+    Query projQuery = new Query("Project");
+
+    Filter ownEditFilter = new CompositeFilter(
```
charmychancey, line 86: No change request here, but can you explain to me what this line does?

dtjanaka: It's creating a new filter for the Datastore query. It's searching for projects with the given ID and with the User as either owner or editor.

src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
   public void doPost(HttpServletRequest request, HttpServletResponse response)
       throws IOException {
     UserService userService = UserServiceFactory.getUserService();
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
 
+    // Must be logged in
     if (!userService.isUserLoggedIn()) {
-      response.sendRedirect("/imgupload.html");
+      throw new IOException("User must be logged in.");
```
robotsaurus, line 48: what catches this exception?  what will the user see if this happens? is this never supposed to happen?  If this is user facing then another option would be to redirect them to the login page.

Instead of asking this every time you throw an exception I am only going to ask it here.  If there are changes to be made then I would do it for every exception.

src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
   public void doPost(HttpServletRequest request, HttpServletResponse response)
```
robotsaurus, line 41: This looks like one very large function. I'd suggest to break it up into smaller pieces to make it easier to read and follow.

src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
...
+          new FilterPredicate("name", FilterOperator.EQUAL, imgName);
+      imgQuery.setFilter(imgFilter);
+      PreparedQuery existingImg = datastore.prepare(imgQuery);
+
+      if (existingImg.countEntities() == 0) {
+        throw new IOException(
+            "No image with that name exists to update for this project or this project does not exist.");
+      }
+      imgEntity = existingImg.asSingleEntity();
+    }
```
robotsaurus, line 149: This set of logic is difficult to understand the intention.  I am wondering why you switch on IsMask and mode == Update.  Why not if on only mode (create/update)?  Why don't we handle not mask and create?

src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
...
+          "User does not have permission to edit the project.");
+    }
+
+    // Asset to update must already exist
+    if (isMask) {
+      Query imgQuery = new Query("Image").setAncestor(projKey);
+      Filter imgFilter =
+          new FilterPredicate("name", FilterOperator.EQUAL, parentImg);
+      imgQuery.setFilter(imgFilter);
+      PreparedQuery existingImg = datastore.prepare(imgQuery);
```
robotsaurus, line 112: This is repeated below so it would be good to break it out into a helper function.

src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
...
 
     BlobstoreService blobstoreService =
         BlobstoreServiceFactory.getBlobstoreService();
     Map<String, List<BlobKey>> blobs = blobstoreService.getUploads(request);
     List<BlobKey> blobKeys = blobs.get("image");
 
     // User submitted form without selecting a file
-    if (blobKeys == null || blobKeys.isEmpty()) {
-      return;
+    Boolean hasNonEmptyImage = !(blobKeys == null || blobKeys.isEmpty());
```
robotsaurus, line 178: I'd suggest the non negative form.

`blobKeys != null && !blobKeys.IsEmpty()`

dtjanaka: Noted

src/main/java/com/google/sps/servlets/BlobServlet.java
dtjanaka: Noted

src/main/java/com/google/sps/servlets/BlobServlet.java
dtjanaka: isMask && create is the default mode which is why no specific block addresses it. I will refactor this logic and add more comments.

src/main/java/com/google/sps/servlets/BlobServlet.java
dtjanaka: Noted

src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
...
 
-    Entity imageEntity = new Entity("Image", userEntity.getKey());
-    imageEntity.setProperty("blobkey", blobKey.getKeyString());
-    imageEntity.setProperty("name", name);
-    datastore.put(imageEntity);
+        if (existingName.countEntities() != 0) {
+          newName += now;
+        }
+        imgEntity.setProperty("name", newName);
+      }
```
robotsaurus, line 236: Potential function.

src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
...
-    query.setFilter(propertyFilter);
+    String now = Instant.now().toString();
+    imgEntity.setProperty("name", imgName);
+    imgEntity.setProperty("utc", now);
+    projEntity.setProperty("utc", now);
 
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    PreparedQuery user = datastore.prepare(query);
+    String tags = request.getParameter("tags");
+    String newName = request.getParameter("new-name");
```
robotsaurus, line 213: These can be moved inside the If statement

src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
...
-      return;
+      // Check validity of blob key
+      BlobKey blobKey = blobKeys.get(0);
+      BlobInfo blobInfo = new BlobInfoFactory().loadBlobInfo(blobKey);
+      String[] splitFilename = blobInfo.getFilename().split(".");
+      String extension = splitFilename[splitFilename.length - 1].toLowerCase();
+      if (blobInfo.getSize() == 0 || !validExtensions.contains(extension)) {
+        blobstoreService.delete(blobKey);
+        throw new IOException("Blobkeys invalid or file not supported.");
+      }
```
robotsaurus, line 202: this can be a function which checks whether the file valid.

src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
...
+        throw new IOException("User submitted form without selecting a file.");
+      }
+    } else if (mode == "update" && !isMask) {
+      blobstoreService.delete(blobKeys.get(0));
+      throw new IOException("No upload allowed for base image update.");
     }
 
-    // Our form only contains a single file input
-    BlobKey blobKey = blobKeys.get(0);
+    if (mode == "create" || (mode == "update" && hasNonEmptyImage)) {
```
robotsaurus, line 188: I think this if condition can be simplified to `if (hasNonEmptyImage)`

src/main/java/com/google/sps/servlets/BlobServlet.java
```diff
@@ -33,59 +41,203 @@
...
+    PreparedQuery accessibleProjects = datastore.prepare(projQuery);
+
+    // Only owners and editors of given project can modify or create
+    if (accessibleProjects.countEntities() == 0) {
+      throw new IOException(
+          "User does not have permission to edit the project.");
+    }
+
+    // Asset to update must already exist
+    if (isMask) {
```
robotsaurus, line 107: this if() {...} part seems like a potential function candidate

charmychancey:
Are we supposed to review this branch anymore? Since you Pr'ed the other ones

PR #11, opened 2020-08-05
add img for mask and scale img

shmcaffrey:


src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -33,31 +42,61 @@ export class EditorComponent implements OnInit {
...
 
-    //adjust canvas to image width and height
-    this.canvas.nativeElement.width = this.image.width;
-    this.canvas.nativeElement.height = this.image.height;
+    //init hidden canvas width to original images width and height
+    this.hiddenCanvas.nativeElement.width = this.image.width;
+    this.hiddenCanvas.nativeElement.height = this.image.height;
+
+    // used to scale the image to the window size, 1.2 so the image is smaller than the window.
+    this.scaleFactor = Math.floor(this.innerHeight / this.image.height / this.scaleFactor);
```
dtjanaka, line 74: It might be more intuitive to have a scale factor < 1 to mean the image is smaller than the window. So instead of dividing by 1.2, multiplying by 0.83?

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,140 @@
...
+  onClick(e: MouseEvent) {
+    const xCoord = e.offsetX;
+    const yCoord = e.offsetY;
+    this.ctx = this.canvas.nativeElement.getContext('2d');
+    let imgData = this.ctx.getImageData(0,0, this.canvas.nativeElement.width, this.canvas.nativeElement.height);
+
+    // tolerance will come from user input
+    const tolerance = 20;
+
+    // returns an array indicies of each pixel in the mask
```
dtjanaka, line 29: nit indices

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,140 @@
...
+   */
+  private getX(width: number, pixel: number, y: number): number {
+    return (pixel / 4) - (y * width);
+  }
+
+  /** 
+  * Convert coord [@x, @y] (2-D) to indexing style of DataArray (1-D)
+  * Returns index of start of pixel at @x, @y
+  * (which represents the red attribute of that pixel)
+  * Important: Returns array of number representing indices.
```
dtjanaka, line 134: Slightly confused by this comment since the comment above already describes the return value.

shmcaffrey:
Link to the live server:[server]( https://accio-step-2020.uc.r.appspot.com/home), I think I may not have pulled the new pom.xml file so it just deployed to default

shmcaffrey:
> Are you able to unstage the commits for top-bar.component.html and top-bar.component.ts? I'm currently working on those files for the database-to-UI part, and I found another solution that works for external redirection which I don't think needs the changes you staged.
> 
> (the changes for the .css of top-bar is fine)

I can try, otherwise when we merge just delete the text from my commit

src/main/ngapp/accio/src/app/editor/editor.component.css
```diff
@@ -3,17 +3,19 @@
   position: relative;
   text-align: center;
   top: 200px;
-  padding-right: 5vw;
-  padding-left: 5vw;
+  /* padding-right: 5vw;
+  padding-left: 5vw; */
```
charmychancey, line 7: Is this intentionally commented out?

src/main/ngapp/accio/src/app/editor/editor.component.css
```diff
@@ -3,17 +3,19 @@
...
   padding-bottom: 100px;
+  padding-top: 100px;
+  background-color: gray;
 }
 
 canvas {
-  width: 100%;
-  height: 100%;
+  /* position: absolute; */
```
charmychancey, line 14: Same here

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,140 @@
...
+    const xCoord = e.offsetX;
+    const yCoord = e.offsetY;
+    this.ctx = this.canvas.nativeElement.getContext('2d');
+    let imgData = this.ctx.getImageData(0,0, this.canvas.nativeElement.width, this.canvas.nativeElement.height);
+
+    // tolerance will come from user input
+    const tolerance = 20;
+
+    // returns an array indicies of each pixel in the mask
+    const maskPixels = this.magicWandService.floodfill(this.originalData, Math.floor(xCoord / this.scale), Math.floor(yCoord / this.scale), tolerance);
```
charmychancey, line 30: I want to discuss our implementation here. If I understand correctly, you're using the original image as the source data here. That means when the user floodfills, a mask is created, and if they floodfill again, the floodfill algorithm will completely ignore the freshly created mask of from the previous floodfill command. We should talk more thoroughly about how adding to the most current version of the mask will work, to ensure that the user has an intuitive experience. Ie is it advantageous/convenient for the user to be able to floodfill once, then floodfill again, but have us account for the most current mask as part of the image being floodfilled.

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   *  Creates a UI mask to the user by reading in all values in the given maskPixels
+   *  Clears the canvas fisrt so the mask can be saved as a new image and then over-
+   *    lays the mask on top of the scaled image. 
+   *  Adds all pixels to mask by giving alpha value of 255, changing global alpha on 
+   *    maskImage will multiply 255 by globalAlpha = [0,1]
+   *  maskPixels: set of coordinates indexed by red value of pixels to be in mask
+   *  imgData: Uint8clampedArray of all pixel data in scaled image. 
+   *  alphaValue: the number the user passes in on a slide bar for transparency of the mask
```
charmychancey, line 44: You can use the '@' symbol as a prefix to denote parameters

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,140 @@
...
+   */
+  private drawMask(maskPixels: Set<number>, imgData: ImageData, alphaValue: number): void {
+    //clear old context
+    this.ctx.clearRect(0,0,this.canvas.nativeElement.width, this.canvas.nativeElement.height);
+    
+    //new imagedata creates transparent black rectangle
+    let img = new ImageData(imgData.width, imgData.height);
+
+    // access all pixels in original mask, scale them up to image on UI width
+    // ***** Need flood fill to pass back total mask, not just new one created, used for undo/redo as well
```
charmychancey, line 54: nit: clarify comment from ***** notation: ie Is this a NOTE or a TODO?

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,140 @@
...
+    newMask.onload = () => {
+        this.ctx.drawImage(newMask, 0, 0);
+    };
+  }
+
+  /**
+  * draws the original image and updated mask on context
+  * imgData: ImageData from the original image
+  * mask: source of new mask 
+  * alphaValue: the number the user passes in on a slide bar for transparency of the mask
```
charmychancey, line 93: Same as prev comment: You can use the '@' symbol as a prefix to denote parameters

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  /** 
+   * Reverses pixel index from floodfill set to get y index
+   * width: width of the original image
+   * pixel: pixel index in from the returned floodfill mask
+   * y: y value from this.getY();
+   */
+  private getX(width: number, pixel: number, y: number): number {
+    return (pixel / 4) - (y * width);
+  }
```
charmychancey, line 128: Just a suggestion, not change req: 
Usually we think of x, y coordinate as being together; does it make more sense to have this as 1 function ie getCoord?

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  /** 
+  * Convert coord [@x, @y] (2-D) to indexing style of DataArray (1-D)
+  * Returns index of start of pixel at @x, @y
+  * (which represents the red attribute of that pixel)
+  * Important: Returns array of number representing indices.
+  * First element contains value of index for attribute: red and so on
```
charmychancey: nit: It might be more clear to denote the order (ie Red, Green, Blue, Alpha) instead of "red and so on"

shmcaffrey:
> > Are you able to unstage the commits for top-bar.component.html and top-bar.component.ts? I'm currently working on those files for the database-to-UI part, and I found another solution that works for external redirection which I don't think needs the changes you staged.
> > (the changes for the .css of top-bar is fine)
> 
> I can try, otherwise when we merge just delete the text from my commit, but I'll let you know if I get it

Would it be okay if we just merge them? I am afraid I'll break something

src/main/ngapp/accio/src/app/mask.directive.ts
shmcaffrey: Yes, I agree. I thought when we talked about the undo/redo, the plan was to have the magic wand api store the cumulative set of all pixels that should be in the mask. For example, if, on the image of four colors, the user clicks the top left quad, then those pixels are added to the set of pixels and saved in the master queue or list. Then when the user clicks the top right quad, the new mask will be created for that quad and we can combine the set created before with the new set created then store the combined set, and pass it back to this directive. You're right that currently the image is just the newest mask, but it appears as if the whole image is in the mask because I get the context of the image displayed when the user clicks on the image, which the old mask is technically still on, once we get the master list/queue of all sets created and combined I will change this.ctx to be a reference to the scaled image in editor.component.ts. I probably should have added comments to explain this and will do that now. Let me know what you're thinking/if that makes sense.

src/main/ngapp/accio/src/app/mask.directive.ts
shmcaffrey: This has to do with your prev comment, I will label them all as TODO-Mask

src/main/ngapp/accio/src/app/mask.directive.ts
shmcaffrey: Alternatively, or in addition, I can create the mask image in editor.component, create the Imagedata for it and then pass the data through so that each time a pixel is added we don't have to redraw the entire mask. I would say in addition though because we will still need the queue/set for undo/redo.

src/main/ngapp/accio/src/app/mask.directive.ts
shmcaffrey: I think it could go either way, I just didn't want to have to pack them together to unpack them again when I need them separate.

charmychancey:
> > > Are you able to unstage the commits for top-bar.component.html and top-bar.component.ts? I'm currently working on those files for the database-to-UI part, and I found another solution that works for external redirection which I don't think needs the changes you staged.
> > > (the changes for the .css of top-bar is fine)
> > 
> > 
> > I can try, otherwise when we merge just delete the text from my commit, but I'll let you know if I get it
> 
> Would it be okay if we just merge them? I am afraid I'll break something

Ok, that should be fine

src/main/ngapp/accio/src/app/mask.directive.ts
charmychancey: So after we implement the idea you're referring to, what would be the result of implementing floodfill on this example:

PREMISE:
There is an image with 3 colors--A, B, and C--oriented like a stack. A is adjacent to B is adjacent to C. Like so:
A
B
C
Floodfill has already been called once by clicking on a B pixel, so mask M would cover the image and the user would see:
A
M
C

PROBLEM:
Now supposed C is a color such that the tolerance would allow floodfill to capture B as well as A inside a mask if we use C as the basis for floodfill's tolerance logic. 
If we clicked on C to do another floodfill command, would the result be:
1.
M
M
M
or 2.
A
M
M

shmcaffrey: 2 would be the result to my knowledge. Although that is not what I thought you were talking about. So are what you are saying is you want the mask to be 1 not 2?

charmychancey: No, outcome 2 makes more sense to me. LGTM

dtjanaka: Just to clarify, to get result 1 implies the floodfill can potentially run over all previously masked pixels every time, correct?

charmychancey: @dtjanaka yes, and so I wanted to make sure to distinguish that possibility and the difference

robotsaurus: A couple comments.

- I really liked be able to add to mask with multiple clicks
- I think having the default be to add and then a button to clear the mask would be nice
- Adding all these details in this thread to the design document would be great if they aren't already there.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -33,31 +44,75 @@ export class EditorComponent implements OnInit {
...
     this.image.src = this.url;
 
-    //adjust canvas to image width and height
-    this.canvas.nativeElement.width = this.image.width;
-    this.canvas.nativeElement.height = this.image.height;
+    //init hidden canvas width to original images width and height
+    this.hiddenCanvas.nativeElement.width = this.image.width;
+    this.hiddenCanvas.nativeElement.height = this.image.height;
+
+    // used to scale the image to the window size, 1.2 so the image is smaller than the window.
```
robotsaurus, line 75: It seems like the comment is out of date because there is no 1.2 below.

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,142 @@
...
+  constructor(
+    private canvas: ElementRef<HTMLCanvasElement>, 
+    private magicWandService: MagicWandService,
+  ) { }
+  private ctx: CanvasRenderingContext2D;
+  private maskImage: HTMLImageElement;
+
+  @HostListener('click', ['$event'])
+  onClick(e: MouseEvent) {
+    const xCoord = e.offsetX;
```
robotsaurus, line 23: Since typescript supports types you'll want to use them.

`const xCoord: number = e.offsetX;`

Here and below.

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,142 @@
...
+    const yCoord = e.offsetY;
+    this.ctx = this.canvas.nativeElement.getContext('2d');
+
+    // tolerance will come from user input
+    const tolerance = 20;
+
+    //  TODO-MASK-maskPixels needs to contain cumulative set of all pixels user clicked and added to the mask
+    // returns an array indices of each pixel in the mask
+    const maskPixels = this.magicWandService.floodfill(this.originalData, Math.floor(xCoord / this.scale), Math.floor(yCoord / this.scale), tolerance);
```
robotsaurus, line 32: Consider having a function that will convert between canvas coordinates and image coordinates.  This way you can implement it once and use it every where.  Additionally, if you need to update it to handle panning, etc. you can easily find all the places you need to update your code to handle this.

src/main/ngapp/accio/src/app/mask.directive.ts
robotsaurus: Something I just realized.  In order to get 2, this means that the floodfill needs to know to stop at previously masked pixels.  I think that seems to work in the current implementation because the clicked on color is not in tolerance of black.  If it was then it would add the masked pixels into the mask.

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    // alpha value hardcoded for now
+    this.drawMask(maskPixels, this.scaledData, 1);
+  }
+
+  /**
+   *  Creates a UI mask to the user by reading in all values in the given maskPixels
+   *  Clears the canvas fisrt so the mask can be saved as a new image and then over-
```
robotsaurus, line 40: nit: first

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -33,31 +44,75 @@ export class EditorComponent implements OnInit {
...
+    let scaleWidth = this.image.width * this.scaleFactor;
+    let scaleHeight = this.image.height * this.scaleFactor;
+
+    //adjust canvas to scaled image width and height, use ctx. 
+    this.canvas.nativeElement.width = scaleWidth;
+    this.canvas.nativeElement.height = scaleHeight;
+    this.ctx.scale(this.scaleFactor, this.scaleFactor); 
+
+    //  initalize transparent black image data to use for mask
+    this.maskData = new ImageData(scaleWidth,  scaleHeight);
```
robotsaurus, line 91: We should make the maskData to be the same size as the original image and then we can use the canvas to scale this just like the image.

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    //  reset global alpha if image was tinted
+    this.ctx.globalAlpha = 1;
+
+    // access all pixels in original mask, scale them up to image on UI width
+    for (let pixel of maskPixels) {
+      let y = this.getY(this.originalData.width, pixel);
+      let x = this.getX(this.originalData.width, pixel, y);
+      this.maskData.data[this.getPixel(x, y, scaledData.width) + 3] = 255;
+    }
```
robotsaurus, line 60: Is this actually scaling the mask up?  So if the image is 1x1 and the scaled up canvas is 2x2 we need to fill all 4 pixels with the mask.  But I think this just does 1 pixel in the scaled mask.

Instead of trying to manually scale the mask up or down can we instead have the mask the same size as the original image and then have canvas scale it?  This is how you do it for the original image.

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,142 @@
...
+  */
+  private drawImageAndMask(scaledData: ImageData, maskUrl: string, alphaValue: number): void {
+    // restore original image (scaledData)
+    this.ctx.putImageData(scaledData, 0, 0);
+
+    //make a new image to add
+    let maskImage = new Image();
+
+    //TODO: alpha change causes old mask to get darker as new pixels added
+    //this.ctx.globalAlpha = alphaValue;
```
robotsaurus, line 105: One possible explanation is if you get the mask pixels from the canvas over time.  This would most likely incorporate the globalAlpha which would explain why adding more pixels the older pixels get darker and darker.

src/main/ngapp/accio/src/app/mask.directive.ts
shmcaffrey: I added these to the design doc under UI and alternative ideas. When you say to clear the mask, do you mean remove the entire mask the user just created? I think that makes sense to have I just wanted to check.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -33,31 +44,75 @@ export class EditorComponent implements OnInit {
...
 
-    //adjust canvas to image width and height
-    this.canvas.nativeElement.width = this.image.width;
-    this.canvas.nativeElement.height = this.image.height;
+    //init hidden canvas width to original images width and height
+    this.hiddenCanvas.nativeElement.width = this.image.width;
+    this.hiddenCanvas.nativeElement.height = this.image.height;
+
+    // used to scale the image to the window size, 1.2 so the image is smaller than the window.
+    this.scaleFactor = Math.floor(this.innerHeight / this.image.height * this.scaleFactor);
```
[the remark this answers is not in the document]
shmcaffrey: I did, the equation is only called once each time the editor component is initialized.

src/main/ngapp/accio/src/app/mask.directive.ts
shmcaffrey: I thought in the style guide in TS, when you initialize the variables when they are declared you didn't need to define the types?

src/main/ngapp/accio/src/app/mask.directive.ts
shmcaffrey: I thought the same this as well (about it not scaling the mask) and originally I had a loop going through all the pixels to set the scaledXscaled array based on the pixel of maskPixels. But then when I created masks, the mask hung over part of the image not in the mask. The bottom and the right sides of the new mask were too large. I believe when I get the new pixel with the scaled width it scales the pixel to the image shown. 

Instead of scaling them 1:1 I agree that scaling them all together like I did with the mask makes sense. I will change it.

shmcaffrey: The ctx that the mask is drawn on is already scaled from editor initialization in editor.ts, I believe that's why the mask hung over the right and bottom sides.

src/main/ngapp/accio/src/app/mask.directive.ts
shmcaffrey: But if I reset the global alpha to 1 before I get that pixels and then reset them all to the alpha value shouldn't this fix the issue of them getting darker and darker because they all start at the same value?

shmcaffrey:
ptal

src/main/ngapp/accio/src/app/mask.directive.ts
charmychancey: @robotsaurus I think the current call to floodfill uses the original ImageData for reference and is unaware of any changes to the working canvas/ImageData, and so it wouldn't "detect" the black. However it is an easy fix, we would just need to pass in a different argument for the floodfill's imgData param

robotsaurus: @charmychancey If we pass in the original data then it's unclear to me how we would get 2 instead of 1 under the current implementation.  Were you suggesting to modify the floodfill algorithm to achieve 2?

robotsaurus: @shmcaffrey Yes I was referring to clear the entire mask.  So start over.  I guess this could be achieved by refreshing the page so probably not a huge priority to implement.

src/main/ngapp/accio/src/app/mask.directive.ts
robotsaurus: Oh you're right. I just looked at the style guide.

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,110 @@
...
+   */
+  private drawMask(maskPixels: Set<number>, scaledData: ImageData, alphaValue: number): void {
+    this.ctx.clearRect(0,0,this.canvas.nativeElement.width, this.canvas.nativeElement.height);
+
+    //  Reset global alpha if image was tinted:
+    //    The max alpha value a pixel can hold is 255*alpha value
+    //    when initialized, alpha value must start at 1 to set properly. 
+    this.ctx.globalAlpha = 1;
+
+    // Access all pixels in original mask and add alpha value so their
```
robotsaurus, line 56: nit: Not a complete sentence.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -10,54 +10,112 @@ import { imageUrls } from '../images';
...
+  /** Returns black transparent ImageData for single mask image. */
+  public getMaskData(): ImageData {
+    return this.maskData;
+  }
+
+  /** Returns the scaled images data for reference printing scaled image after mask. */
+  public getScaledData(): ImageData {
+    return this.ctx.getImageData(0, 0, this.image.width * this.scaleFactor, this.image.height * this.scaleFactor);
+  }
 }
```
robotsaurus, line 121: I might be missing something but doesn't this return the scaled original image data PLUS the scaled mask?  Since we are getting the image data from the main canvas?  I think it could also be the cumulative mask from multiple user clicks.

If this is correct, then I think the name originalData, scaledData and maskData are a bit confusing.  I assumed scaledData was the originalData just scaled to the size of the canvas.  

I think we should clarify the names a bit. Data is a bit of a generic word.  If possible leave it out.  One suggestion is (but it may not be perfect):

originalData -> originalImage
scaledData - > scaledImageAndMask
maskData -> mask

This function name is probably better called `getScaledImgAndMask`.

src/main/ngapp/accio/src/app/mask.directive.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    // TODO(shmcaffrey): Tolerance will come from user input.
+    const tolerance = 20;
+
+    //  TODO-MASK-maskPixels needs to contain cumulative set of all pixels user clicked and added to the mask.
+    // returns an array indices of each pixel in the mask.
+    const maskPixels = this.magicWandService.floodfill(this.originalData, Math.floor(xCoord / this.scale), Math.floor(yCoord / this.scale), tolerance);
+
+    // TODO(shmcaffrey): change Alpha value to incorperate user input.
+    this.drawMask(maskPixels, this.scaledData, 1);
```
robotsaurus, line 35: Is there a reason you are passing in the scaledData instead of the originalData and having canvas scale when drawing?  Is is for performance?

src/main/ngapp/accio/src/app/editor/editor.component.ts
shmcaffrey: the scaled mask is an object of type imageData but it's not drawn on any canvas, and if it was it would be transparent and black so it wouldn't be seen. And yes scaledData is just originalData but scaled to the size of the canvas. 

If it's okay, I will update the names for the next pr because I forgot to make a separate branch for the toolbar and I don't want to prolong the pr by adding more files to it which would be necessary to push the name changes and have the pr in working condition.

src/main/ngapp/accio/src/app/mask.directive.ts
shmcaffrey: Noted in chat

shmcaffrey:
ptal

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -10,54 +10,112 @@ import { imageUrls } from '../images';
...
-    this.canvas.nativeElement.width = this.image.width;
-    this.canvas.nativeElement.height = this.image.height;
+    let imgWidth = this.image.width;
+    let imgHeight = this.image.height;
+
+    // Initialize hidden canvas width and height to original images width and height.
+    this.hiddenCanvas.nativeElement.width = imgWidth;
+    this.hiddenCanvas.nativeElement.height = imgHeight;
+
+    // Used to scale the image to the window size, 1.2 so the image is smaller than the window.
```
robotsaurus, line 81: nit: I don't know where the 1.2 is coming from in this comment.  It seems like might be a stale comment.  If not then maybe expand upon where the 1.2 is coming from.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -10,54 +10,112 @@ import { imageUrls } from '../images';
...
+    this.scaleFactor = Math.floor(this.innerHeight / imgHeight * this.scaleFactor);
+    // TODO(shmcaffrey): add scaling if image is larger than window
+    if (this.scaleFactor <= 0) {
+      this.scaleFactor =  1;
+    }
+
+    // Adjust canvas to scaled image width and height, use ctx. 
+    this.canvas.nativeElement.width = imgWidth* this.scaleFactor;
```
robotsaurus, line 89: nit (style guide): space between imgWidth and *

src/main/ngapp/accio/src/app/editor/editor.component.ts
robotsaurus: Sounds good to me.  You can resolve the conversation so it doesn't block you.

We should also chat over VC because I am still confused about what getImageData returns.

src/main/ngapp/accio/src/app/editor/editor.component.ts
shmcaffrey: I also changed this comment?? I don't know why those changes never got pushed, I changed it to .9, the value of scaleFactor. This is just a factor to make the scaled image smaller than the users screen, .9/1.2 are both arbitrary, they could be any number.

PR #12, opened 2020-08-06
Only project servlet code.

dtjanaka:


src/main/java/com/google/sps/servlets/ProjectServlet.java
```diff
@@ -0,0 +1,266 @@
...
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
+    UserService userService = UserServiceFactory.getUserService();
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    // Must be logged in
+    if (!userService.isUserLoggedIn()) {
+      response.sendRedirect("/"); // placeholder: should redirect to login
```
shmcaffrey, line 46: I forget if we are just using the gmail api for log in or something else. Basically just wondering if this should redirect to the gmail api or our 'log in' page/homepage?

src/main/java/com/google/sps/servlets/ProjectServlet.java
```diff
@@ -0,0 +1,266 @@
...
+    Entity projEntity = new Entity(DataUtils.PROJECT);
+
+    if (!isCreateMode) {
+      // Must be owner to update
+      projEntity = DataUtils.getProjectEntity(projId, uEmail, false, false);
+
+      // Delete overrides all other updates
+      boolean delete = Boolean.parseBoolean(request.getParameter("delete"));
+      if (delete) {
+        datastore.delete(projEntity.getKey());
```
shmcaffrey, line 67: We should also send the user an alert to make sure they actually want to delete the proj, always good to have a safe guard, this would be in the UI though when they click the button, I just wanted to note it and add it to the design doc.

src/main/java/com/google/sps/servlets/ProjectServlet.java
```diff
@@ -0,0 +1,266 @@
...
+      }
+      // Don't do anything if updating and no name provided
+    }
+
+    // Default visibility to private (only owners and editors can view)
+    String visibility = request.getParameter("visibility");
+    if (DataUtils.isEmptyParameter(visibility) && isCreateMode) {
+      visibility = DataUtils.PRIVATE;
+    }
+    if (!DataUtils.isEmptyParameter(visibility) &&
```
shmcaffrey, line 95: else if?

src/main/java/com/google/sps/servlets/ProjectServlet.java
```diff
@@ -0,0 +1,266 @@
...
+      }
+      listOwnerEmails.add(uEmail);
+      projEntity.setIndexedProperty(
+          "owners", DataUtils.withDuplicatesRemoved(listOwnerEmails));
+    }
+
+    // If anything provided for editors, overwrite current editors
+    if (!DataUtils.isEmptyParameter(editorsString)) {
+      ArrayList<String> listEditorEmails =
+          DataUtils.parseCommaList(editorsString);
```
shmcaffrey, line 123: What if the user enters the list with semi-colons seperating, or just spaces?

src/main/java/com/google/sps/servlets/ProjectServlet.java
dtjanaka: Right, since we're using the Users API, there is no login page, so I have it redirect to the homepage for now. Is there a better place to redirect to?

src/main/java/com/google/sps/servlets/ProjectServlet.java
dtjanaka: Both statements may be run, which is why there are two if statements. The first one handles the default visibility for create mode and the second one actually sets the entity property.

src/main/java/com/google/sps/servlets/ProjectServlet.java
dtjanaka: It would not work with the current implementation. Unsure how semicolons vs commas vs spaces would be differentiated, especially if spaces could be used within tags.

src/main/java/com/google/sps/servlets/ProjectServlet.java
charmychancey: The plan is to have an IntroComponent eventually, so /intro when that time comes

src/main/java/com/google/sps/servlets/ProjectServlet.java
```diff
@@ -0,0 +1,284 @@
...
+
+    // Default visibility to private (only owners and editors can view)
+    String visibility = request.getParameter("visibility");
+    if (DataUtils.isEmptyParameter(visibility) && isCreateMode) {
+      visibility = DataUtils.PRIVATE;
+    }
+    if (!DataUtils.isEmptyParameter(visibility) &&
+        (visibility.toLowerCase().equals(DataUtils.PUBLIC) ||
+         visibility.toLowerCase().equals(DataUtils.PRIVATE))) {
+      projEntity.setProperty("visibility", visibility);
```
snazzorama, line 105: Might be better to store `visibility.toLowerCase()`, so that the values in your database are standardized.

PR #14, opened 2020-08-06
Only login code.

dtjanaka:


src/main/java/com/google/sps/servlets/Login.java
```diff
@@ -0,0 +1,38 @@
...
+ * Sends back the login status of User and a login/logout link.
+ */
+@WebServlet("/login-status")
+public class Login extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
+    response.setContentType("application/json");
+
+    String page = request.getParameter("page");
```
shmcaffrey, line 23: Can you explain what page a little? I get that it's the redirection page but where does it come from?

dtjanaka: Since this function handles GET requests, the page parameter comes from the URL itself. A request to this servlet might look like /login-status?page=projects, which would mean the servlet would return a login/logout link that, once clicked, would send the user to the projects page after logging them in or out.

src/main/java/com/google/sps/servlets/Login.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Sends back the login status of User and a login/logout link.
```
charmychancey, line 14: Can you include something like @return {string} for type(login/logout link)

charmychancey: Same with type(login status)

charmychancey: lgtm after that

PR #16, opened 2020-08-06
Helper servlets and classes for blobs servlet.

dtjanaka:


src/main/java/com/google/sps/servlets/BlobUrl.java
```diff
@@ -10,6 +10,9 @@
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+/**
+ * Returns a Blobstore upload link.
```
charmychancey, line 14: Can you include @return {data-type} for type(upload link)

charmychancey: lgtm after that

PR #17, opened 2020-08-08
Database to front end implementation -- projects, login

charmychancey:
Connect UI to database
- Connect login-status servlet; user can log in to access projects. 
    - User is redirected to intro page if not logged in.
    - Logging in redirects the user to their home page.
- Connect projects servlet:
    - User can create a project, which will navigate them to the new project page.
    - User can apply filters and then search for their current projects. This returns a list of their projects to them.
        - Users can click on any project that appears, which navigates them to that project's page (img-gallery/:proj-id)


dtjanaka:
Would it be possible to uncommit the database files so only the front end implementation shows up?

charmychancey:
> Would it be possible to uncommit the database files so only the front end implementation shows up?

I don't think so, at least as far as I am aware.

dtjanaka:
> > Would it be possible to uncommit the database files so only the front end implementation shows up?
> 
> I don't think so, at least as far as I am aware.
> 
> I know there is a feature to revert the branch to a previous commit, which "removes", but that isn't what we would want in this case since that would remove all of my commits too and just restore the branch to the chosen commit.
> 
> However, if what you're worried about is merge conflicts with master, I already merged with master on this branch around mid/late noon on August 7th and then resolved merge conflicts before PRing

I'm more concerned that this PR is dependent on the database code which is likely to change since it is also being PR'ed atm. It would also be easier to review with only the UI code shown.

charmychancey:
@dtjanaka Do you have a preference for how to proceed then? Ie I'm fine with getting this PR reviewed and merged for now, and it it breaks later we can go back to it. Having it initially reviewed for now at least would hopefully make future change requests to the code minimal if it needs to change again after the new servlet code is merged.

Also, I will attach a screenshot below showing where to uncheck the filter so that you only see the ui code. Just uncheck the .java files
github filter snipet

dtjanaka:
@charmychancey I didn't know about the file filter; good to know! I'll review the UI code but maybe we can ask Brett and Mike on Monday.

src/main/ngapp/accio/src/app/home/home.component.ts
```diff
@@ -1,15 +1,49 @@
...
+import { Router } from '@angular/router';
+import * as $ from 'jquery';
 
 @Component({
   selector: 'app-home',
   templateUrl: './home.component.html',
   styleUrls: ['./home.component.css']
 })
 export class HomeComponent implements OnInit {
+  // TODO: add filter global once talk w/ DB team to understand it.
```
dtjanaka, line 11: Noted

src/main/ngapp/accio/src/app/top-bar/top-bar.component.css
```diff
@@ -1,11 +1,10 @@
 #header {
```
dtjanaka, line 1: nit: 2 spaces instead of 4

src/main/ngapp/accio/src/app/top-bar/top-bar.component.html
```diff
@@ -1,7 +1,7 @@
 <div id="header">
-  <button routerLink="/img-gallery" class="login-button">Sign In</button>
+  <button id="login-button" class="login-button">{{ buttonText }}</button>
   
-  <a href="home">
-    <img height="100%" src="assets/images/logo.png">
-  </a>
+    <a href="home">
+        <img height="100%" src="assets/images/logo.png">
```
dtjanaka, line 5: nit: spacing

charmychancey:
> I'm more concerned that this PR is dependent on the database code which is likely to change since it is also being PR'ed atm. It would also be easier to review with only the UI code shown.

@dtjanaka Can you clarify on what might change in the servlets code to affect the UI code? Because the UI code side, the UI code is only dependent on what JSON object is being returned by the servlets atm

dtjanaka:
> 
> 
> > I'm more concerned that this PR is dependent on the database code which is likely to change since it is also being PR'ed atm. It would also be easier to review with only the UI code shown.
> 
> @dtjanaka Can you clarify on what might change in the servlets code to affect the UI code? Because on the UI code side, the UI code is only dependent on what JSON object is being returned by the servlets atm

You're right, I worded that vaguely. To clarify what I meant, I think merging this PR would be messy. Correct me if I'm wrong, but it seems merging this to master before the database PRs would mean those PRs would not show all the changes there because they compare to master. But then merging this after would overwrite the edits to the database code not in this branch.

charmychancey:
@dtjanaka ....umm, I think I tried to quote and reply to your comment, but I may have accidentally ended up editting your comment instead of replying to it. Had no idea that was even possible. However, your original comment is still inside YOUR comment that I edited lol. So to make it clear, the comment by dtjanaka above this comment contains both dylon's comment(which is the quoted part) and my comment(which is the normal looking text). Sorry about that.

src/main/ngapp/accio/src/app/home/home.component.ts
```diff
@@ -1,15 +1,67 @@
...
+  }
+  
+  // Redirects user to intro page if not logged in
+  async handleLoggedOut(): Promise<void> {
+    console.log('handling login...');
+    const response = await fetch('/login-status?page=');
+    /**Content received contains 
+     * {loggedIn: boolean,
+     * url: string} 
+     */
```
snazzorama, line 36: I'd suggest not using the `/** ... */` style of comment unless you are making JSDoc comments. For general comments, use `//` or `/* ... */`.

src/main/ngapp/accio/src/app/home/home.component.html
```diff
@@ -1,8 +1,51 @@
...
+    <mat-select [(ngModel)]="filterSort">
+      <mat-option value="dsc">newest to oldest</mat-option>
+      <mat-option value="asc">oldest to newest</mat-option>
+    </mat-select>
+    <mat-hint align="end">Default: newest</mat-hint>
+  </mat-form-field>
+
+  <mat-form-field hintLabel="Clear this field to remove this filter" appearance="fill">
+    <mat-label>Project ID</mat-label>
+    <input matInput [(ngModel)]="filterId" placeholder="Ex. ajkhweAfdaDFsh">
```
snazzorama, line 40: What is that example string?

charmychancey: It is an example of what a project-id can look like. It doesn't actually belong to a project though.

charmychancey:
Ptal

PR #20, opened 2020-08-10
merge back to database-front-end

shmcaffrey:
Note: I will add the ability to save masks to datastore on 8/11


src/main/ngapp/accio/src/app/toolbar/toolbar.component.ts
```diff
@@ -12,4 +12,7 @@ export class ToolbarComponent implements OnInit {
   ngOnInit(): void {
   }
 
+  getTolerance(value: number) {
```
dtjanaka, line 15: What is this for?

shmcaffrey: I put off integrating user input for tolerance in the mask toolabar until the MVP is done (ie being able to upload photos, I will add a comment.

PR #22, opened 2020-08-12
Add checks, add indexes, and fix response format.

dtjanaka:


shmcaffrey:
lgtm

src/main/java/com/google/sps/servlets/BlobHost.java
```diff
@@ -32,17 +41,73 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+      }
+    } else { // if Image, get parent --> Project
+      try {
+        projId = (String)datastore.get(assetEntity.getParent())
+                     .getProperty("proj-id");
+      } catch (Exception e) {
+        throw new IOException(
+            "You do not have permission to access this project.");
+      }
+    }
```
snazzorama, line 76: There's a bunch of duplicated code between the two branches. How about putting the `if` inside of the `try`? Or make a variable for the object whose proj-id you are going to access.

src/main/java/com/google/sps/servlets/BlobHost.java
```diff
@@ -32,17 +41,73 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+
+    String uEmail = userService.getCurrentUser().getEmail();
```
snazzorama, line 78: Try to avoid abbreviations; we can call this `userEmail`.

PR #23, opened 2020-08-12
add user ability to upload images and masks, add toolbar UI

shmcaffrey:
Note that the toolbar doesn't have functionality yet, will be working on that after adding semantic labels.


src/main/ngapp/accio/src/app/editor/editor.component.html
```diff
@@ -1,6 +1,19 @@
...
-  </canvas>
-  <canvas #hiddenCanvas style="display:none">
-  </canvas>
-</div>
\ No newline at end of file
+<form novalidate [formGroup]="uploadMaskForm" (ngSubmit)="onSubmit()"  
+      *ngIf="display" method="POST" enctype="multipart/form-data" id="mask-form">
+  <mat-form-field appearance="fill">
+    <mat-label>Mask Title</mat-label>
+    <input matInput formControlName="maskName" placeholder="Untitled"/>
```
charmychancey, line 5: You can write something like "mask name" in placeholder, to label the input when the user hasn't clicked on the field yet.

src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.ts
```diff
@@ -1,16 +1,115 @@
...
+  ) { }
 
   ngOnInit(): void {
+    this.uploadImageForm = new FormGroup({
+      imgName: new FormControl(),
+      image: new FormControl(),
+      tags: new FormControl()
+    });
+
+    //creates the form data of parameters to be sent to servlet
```
charmychancey, line 36: nit: consistency in comments; sometimes there is a space after the // and sometimes not, also sometimes there is capitalization on the first char and sometimes not.

src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.ts
```diff
@@ -1,16 +1,115 @@
...
+    this.formData = new FormData();
+
+    // Set the project id, first try if the 
+    //    project is a new project and there are query keys
+    this.route.queryParams.subscribe(params => {
+      this.projectId = params['proj-id'];
+    });
+    //if no keys then just get the url params
```
charmychancey, line 44: nit: consistency in comments; same as above

src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.ts
```diff
@@ -1,16 +1,115 @@
...
+    const imageContent = await response.json();
+    this.imageArray = imageContent;
+
+    if (this.imageArray.length > 0) {
+      this.displayImages = true;
+    }
+  }
+
+  /** 
+   *  @param formData is initalized with values given by user.
```
charmychancey, line 85: I'm not sure if '@param' is the correct semantic here. Might be a bit confusing since it's 'imgName', 'tags', and 'image' that are the actual arguments received from the form. But I'm also not sure what the correct wording is.

src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.ts
```diff
@@ -1,16 +1,115 @@
...
+
+
+    this.formData.append('tags', this.uploadImageForm.get('tags').value);
+    console.log('FORM DATA tags ' + this.uploadImageForm.get('tags').value);
+
+    this.formData.append('proj-id', this.projectId);
+    console.log('FORM DATA proj-id ' + this.projectId);
+
+    this.postBlobsService.onUpload(this.formData, this.uploadImageForm);
+    // upload reset and call page refresh
```
charmychancey, line 113: nit: consistency between comments; same as above

src/main/ngapp/accio/src/app/toolbar/toolbar.component.html
```diff
@@ -1,3 +1,20 @@
-<div id="toolbar">
-this is div
-</div>
\ No newline at end of file
+<mat-button-toggle-group vertical style="border:none">
+  <mat-button-toggle mat-icon-button value="magic-wand" title="Magic Want">
```
charmychancey, line 2: typo: title="Magic Wan**d**"

charmychancey:
lgtm

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -27,95 +37,161 @@ export class EditorComponent implements OnInit {
...
+
+    this.formData = new FormData();
+  }
+
+ /** 
+  *  Builds ImageBlob to be appended to form and posted.
+  *    projectIdIn: string, 
+  *    imageNameIn: string, 
+  *    modeIn: string, 
+  *    imageIn: any = '',
```
snazzorama, line 173: It's best to not use the `any` type in TypeScript. Do we know what type this thing is going to be? If so, you should make it have that type. If you want to denote that nothing is stored, it would be better to make it nullable, e.g.

`imageIn: Foo | null = null,`

src/main/ngapp/accio/src/app/editor/mask.directive.ts
```diff
@@ -60,13 +60,16 @@ export class MaskDirective {
     //    when initialized, alpha value must start at 1 to set properly. 
     this.ctx.globalAlpha = 1;
 
-    // Access all pixels in the original mask and add alpha value so they're visable.
+    //  Access all pixels in the original mask and add alpha value so they're visable.
```
snazzorama, line 63: visible, not visable

src/main/ngapp/accio/src/app/editor/editor.component.ts
shmcaffrey: Technically, when the user uploads an image from their computer the the type is a file and then when they upload the mask it is a blob type. So I didn't want to limit the type. But I tried it with type Blob and it works so I will change that.

src/main/ngapp/accio/src/app/ImageBlob.ts
```diff
@@ -0,0 +1,40 @@
...
+export class ImageBlob {
+  // Required params
+  projectId: string;
+  imageName: string;
+  mode: string;
+
+  image: any;
+  parentImageName: string;
+  newImageName: string;
+  tags: string;
```
robotsaurus, line 17: Are these expected to be comma separated?  Something like this is good to clarify so people knew to the codebase can find out.

src/main/ngapp/accio/src/app/editor/editor.component.ts
robotsaurus: If you need multiple types then I think you can use a type union.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -27,95 +37,161 @@ export class EditorComponent implements OnInit {
...
+      maskName: new FormControl(),
+      labels: new FormControl(),
+    });
+
+    this.formData = new FormData();
+  }
+
+ /** 
+  *  Builds ImageBlob to be appended to form and posted.
+  *    projectIdIn: string,
```
robotsaurus, line 170: this comment seems out of place.  Not sure what I am supposed to get out of this.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -27,95 +37,161 @@ export class EditorComponent implements OnInit {
...
   public getScaledData(): ImageData {
     return this.ctx.getImageData(0, 0, this.image.width * this.scaleFactor, this.image.height * this.scaleFactor);
   }
+
+  /** 
+   *  Clears canvas to get mask Image as url to store and redraws
+   *    original image for possibility user makes more edits. 
+   * @return Url for the mask image to be stored in blobstore.  
+   */
+  async getMaskBlob(): Promise<void> {
```
robotsaurus, line 140: Since you are using the hiddenCtx in both the getImgData function and the getMaskBlob function there is the potential for a race condition and conflict.  How are you avoiding that?  It would be good to clarify that in comments in the code.

shmcaffrey: I noted this in the comments, I am going to re factor mask.directive and editor.ts today so I will account for this then.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -27,95 +37,172 @@ export class EditorComponent implements OnInit {
...
+
+    let imageBlob = new ImageBlob(
+      this.projectId, 
+      this.uploadMaskForm.get('maskName').value,
+      'create', 
+      this.blobMask,
+      this.parentName,
+      '', '',
+      this.uploadMaskForm.get('labels').value
+    );
```
robotsaurus, line 201: I understand the reasoning why you are adding the function comment with the ImageBlob.  It is hard to remember the order of the parameters of the constructor.  Instead of duplicating the comments for every function that calls ImageBlob it is better to make sure the ImageBlob calls are self-documented. This is because they are less likely to go stale and out of date.  So for most things you are using good variable names.  Where you aren't then I would add a comment that documents them.  So something like:

```
    let imageBlob = new ImageBlob(
      this.projectId, 
      /*imageName=*/this.uploadMaskForm.get('maskName').value,
      /*mode=*/'create', 
      this.blobMask,
      this.parentName,
      /*newImageName=*/'',
      /*tags=*/ '',
      /*delete=*/this.uploadMaskForm.get('labels').value
    );
```

src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.ts
```diff
@@ -1,16 +1,126 @@
...
+  *      if a value later in the constructor is applicable.
+  *
+  *         projectIdIn: string, 
+  *         imageNameIn: string, 
+  *         modeIn: string, 
+  *         imageIn: any = '',
+  *         parentImageNameIn: string = '',
+  *         newImageNameIn: string = '',
+  *         tagsIn: string = '',
+  *         deleteIn: string = 'delete'
```
robotsaurus, line 98: Same thing about making the callsite self documenting.

src/main/ngapp/accio/src/app/ImageBlob.ts
```diff
@@ -0,0 +1,55 @@
...
+  // default parameters on non-required parameters 
+  constructor(
+    projectIdIn: string, 
+    imageNameIn: string, 
+    modeIn: string, 
+    imageIn: Blob,
+    parentImageNameIn: string = '',
+    newImageNameIn: string = '',
+    tagsIn: string = '',
+    deleteIn: string = 'false'
```
robotsaurus, line 44: It is typically better to store booleans as booleans instead of strings.  Does it add a lot of complications if you use a boolean here?

shmcaffrey: When appending delete to the form, it wouldn't append a bool, so I changed it to a string, but I will use the toString method so there is no confusion when initializing an ImageBlob

PR #25, opened 2020-08-12
Implement fetching masks on img-gallery component

charmychancey:
Implement fetching masks for img-gallery component:
- Add a search bar with filter parameters to ImgGalleryComponent
- Add 'show-mask' filter to fetch masks from database

### DIFF
--- a/src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.html
+++ b/src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.html
@@ -23,10 +23,74 @@
   <button type="submit" mat-button  id="upload-button">Upload</button>
 </form>
 
-<div id="gallery" *ngIf="displayImages">
+<br>
+<br>
+<br>
+<br>
+<br>
+<br>
+<br>
+<br>
+<br>
+
+<!--Filter Bar-->
+<h1>Search by:</h1>
+<div class="container">
+  <mat-form-field hintLabel="Clear to remove this filter" appearance="fill">
+    <mat-label>Image Name</mat-label>
+    <input matInput [(ngModel)]="imgName" placeholder="Ex. MyImage">
+  </mat-form-field>
+
+  <mat-form-field hintLabel="Clear to remove this filter" appearance="fill">
+    <mat-label>Mask Name</mat-label>
+    <input matInput [(ngModel)]="maskName" placeholder="Ex. MyMask">
+  </mat-form-field>
+
+  <mat-form-field hintLabel="Clear to remove this filter" appearance="fill">
+    <mat-label>Filter Tag</mat-label>
+    <input matInput [(ngModel)]="tag" placeholder="Ex. river">
+  </mat-form-field>
+
+  <mat-form-field appearance="fill">
+    <mat-label>Sort Images</mat-label>
+    <mat-select [(ngModel)]="sortImg">
+      <mat-option value="dsc">newest to oldest</mat-option>
+      <mat-option value="asc">oldest to newest</mat-option>
+    </mat-select>
+    <mat-hint align="end">Default: newest</mat-hint>
+  </mat-form-field>
+
+  <mat-form-field appearance="fill">
+    <mat-label>Show masks for Image: {{ imgName }}</mat-label>
+    <mat-select [(ngModel)]="withMasks">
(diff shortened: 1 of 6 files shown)


src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.html
```diff
@@ -23,10 +23,74 @@
   <button type="submit" mat-button  id="upload-button">Upload</button>
 </form>
 
-<div id="gallery" *ngIf="displayImages">
+<br>
```
charmychancey, line 26: Note: mat-forms being difficult with 'display: relative' here. UI team set the css up for the html, so the <br> is just a placeholder for when UI team wants to correct the display placement of the search bar against the upload bar

charmychancey, line 26: The forms was being difficult with overlapping, I suspect "display: relative". Didn't want to change css scheme that UI team had set with consistency of display attribute, so `<br>` can just be a placeholder if UI team want's to change this overlap correction with css.

charmychancey, line 26: NOTE: The forms was being difficult with overlapping, I suspect "display: relative". Didn't want to change css scheme that the UI team had set with consistency/uniformity of the display attribute, so <br> can just be a placeholder if UI team wants to change this overlap correction with css later on when revisiting this directory to implement img download.

dtjanaka: Instead of these `<br>` s, maybe set the bottom margin of the element above?

charmychancey: Noted. Messing with properties of element above gave undesirable css outcomes to the background color. Instead, I will remove the `<br>` and alter the css for the elements below, which have no prior css formatting as they are code introduced in my commit.

src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.html
```diff
@@ -23,10 +23,74 @@
...
     </a>
   </div>
-</div>
\ No newline at end of file
+</div>
+
+<div class="gallery">
+  <ng-container *ngFor="let image of imageArray">
+    <div *ngFor="let mask of imageArray[0]['masks']" class="gallery-item">
+      <a [routerLink]="['/editor', projectId, mask['name'], mask['url']]">
```
shmcaffrey, line 79: If the user just opens up the mask, we should have a way to show the parent image on the editor behind it, possibly pass the image url and mask url and if the mask url is null then we just draw the image, but if it's not null then draw the mask over top. I will implement this in editor when refactoring if you want to adjust the router link

src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.html
shmcaffrey: I will take a look at changing relative. I think it just needs to be consistent over the application. I can also make a break div which spaces the two apart.

src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.html
```diff
@@ -23,10 +23,74 @@
...
+      <p>{{ image['name'] }}</p>
     </a>
   </div>
-</div>
\ No newline at end of file
+</div>
+
+<div class="gallery">
+  <ng-container *ngFor="let image of imageArray">
+    <div *ngFor="let mask of imageArray[0]['masks']" class="gallery-item">
```
shmcaffrey, line 78: imageArray only contains one image in this case right? Could you say "*ngFor="let mask of image['masks']"

charmychancey: No, because the image does not exist until the filter has been submitted by the user(has not been fetched yet). Therefore, when I have an ngFor that tries to access image['masks'] I get undefined errors. That is why I need to reference image as an element of imageArray first. Basically, if there is no image yet, my current code will not execute on any element, whereas if I changed to your suggestion, it would initially try to run on an undefined attribute.

charmychancey: That should be fine

PR #29, opened 2020-08-14
MaskController class skeleton, SetOperator class

charmychancey:
- Set up MaskController class skeleton. The actual code within the MaskController methods are subject to change, I only set up suggestive code, but @dtjanaka will actually decide on more specific design and implementation for this class.
- Make SetOperator class for basic Set operations because this does not exist within javascript's standard lib


src/main/ngapp/accio/src/app/editor/set-operator.ts
```diff
@@ -0,0 +1,52 @@
+// Class to perform basic set operations
+export class SetOperator {
+  // Ie. [1, 2, 3] union [3, 4, 5] ==> [1, 2, 3, 4, 5]
```
dtjanaka, line 3: Nit: ie = that is, perhaps eg would be better

PR #34, opened 2020-08-17
add maskTool  enum, fix maskTool init

shmcaffrey:
small pr to fix bug where maskTool wasn't initialized when editor opened.

### DIFF
--- a/src/main/ngapp/accio/src/app/editor/MaskToolEnum.ts
+++ b/src/main/ngapp/accio/src/app/editor/MaskToolEnum.ts
@@ -1,4 +1,5 @@
 export enum MaskTool {
   magicWand = 'MAGIC-WAND', 
-  maskOnly = 'MASK-ONLY'
+  maskOnly = 'MASK-ONLY',
+  paint = 'PAINT'
 }
--- a/src/main/ngapp/accio/src/app/editor/MaskToolEnum.ts
+++ b/src/main/ngapp/accio/src/app/editor/MaskToolEnum.ts
@@ -1,5 +1,5 @@
 export enum MaskTool {
-  magicWand = 'MAGIC-WAND', 
-  maskOnly = 'MASK-ONLY',
-  paint = 'PAINT'
+  MAGIC_WAND = 'MAGIC-WAND', 
+  MASK_ONLY = 'MASK-ONLY',
+  PAINT = 'PAINT'
 }
--- a/src/main/ngapp/accio/src/app/editor/MaskToolEnum.ts
+++ b/src/main/ngapp/accio/src/app/editor/MaskToolEnum.ts
@@ -1,0 +1,4 @@
+export enum MaskTool {
+  magicWand = 'MAGIC-WAND', 
+  maskOnly = 'MASK-ONLY'
+}
--- a/src/main/ngapp/accio/src/app/editor/editor.component.ts
+++ b/src/main/ngapp/accio/src/app/editor/editor.component.ts
@@ -68,7 +68,7 @@ export class EditorComponent implements OnInit {
     this.tolerance = 30;
     this.maskAlpha = 1;
     this.disableFloodFill = false;
-    this.maskTool = MaskTool.magicWand;
+    this.maskTool = MaskTool.MAGIC_WAND;
     
     this.route.paramMap.subscribe(params => {
       this.projectId = params.get('proj-id ');
@@ -193,7 +193,7 @@ export class EditorComponent implements OnInit {
     let mask = new Image();
     mask.onload = () => {
       this.clearCanvas();
-      if (this.maskTool != MaskTool.maskOnly) {
+      if (this.maskTool != MaskTool.MASK_ONLY) {
         this.drawScaledImage(this.image);
       }
       this.ctx.save();
@@ -278,7 +278,7 @@ export class EditorComponent implements OnInit {
     this.disableFloodFill = true;
     this.maskImageData = new ImageData(this.image.width, this.image.height);
     this.drawMask();
-    if (this.maskTool == MaskTool.magicWand) {
+    if (this.maskTool == MaskTool.MAGIC_WAND) {
       this.disableFloodFill = false;
     }
   }
@@ -300,7 +300,7 @@ export class EditorComponent implements OnInit {
       this.maskImageData.data[i + 3] = 255 - this.maskImageData.data[i+ 3];
     }
     this.drawMask();
-    if (this.maskTool == MaskTool.magicWand) {
+    if (this.maskTool == MaskTool.MAGIC_WAND) {
       this.disableFloodFill = false;
     }
     this.disableSubmit = false;
@@ -322,7 +322,7 @@ export class EditorComponent implements OnInit {
     //  Draw mask with new maskAlpha value.
     this.disableFloodFill = true;
     this.drawMask();
-    if (this.maskTool == MaskTool.magicWand) {
+    if (this.maskTool == MaskTool.MAGIC_WAND) {
       this.disableFloodFill = false;
     }
     console.log('new maskAlpha: ' + value);
@@ -337,14 +337,14 @@ export class EditorComponent implements OnInit {
     this.disableFloodFill = true;
     switch (tool) {
       case 'MAGIC-WAND': 
-        this.maskTool = MaskTool.magicWand;
+        this.maskTool = MaskTool.MAGIC_WAND;
         this.disableFloodFill = false;
         break;
       case 'PAINT':
-        this.maskTool = MaskTool.paint;
+        this.maskTool = MaskTool.PAINT;
         break;
       case 'MASK-ONLY':
-        this.maskTool = MaskTool.maskOnly;
+        this.maskTool = MaskTool.MASK_ONLY;
         break;
     }
     console.log('switched tool to ' + this.maskTool);
@@ -355,7 +355,7 @@ export class EditorComponent implements OnInit {
  /**
   *  Adds pixel user painted to mask.
   *  TODO: Possibly change the implementation so the pixel drawn is
-  *        replecated on the screen ASAP, and then once the user finishes 
+  *        replicated on the screen ASAP, and then once the user finishes 
   *        drawing (mouse up) then the real mask is drawn based on the set.
   *        Would decrease lag.
   */
--- a/src/main/ngapp/accio/src/app/editor/editor.component.ts
+++ b/src/main/ngapp/accio/src/app/editor/editor.component.ts
@@ -5,6 +5,7 @@ import { FormGroup, FormControl } from '@angular/forms';
 
 import { PostBlobsService } from '../post-blobs.service';
 import { ImageBlob } from '../ImageBlob';
+import { MaskTool } from './MaskToolEnum';
 
 @Component({
   selector: 'app-editor',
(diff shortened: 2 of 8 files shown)


src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -158,10 +162,10 @@ export class EditorComponent implements OnInit {
   *  Sets new pixels in magenta, clears canvas of previous data 
   *    and draws image and mask as scaled. Disables submit
   *    on mask until the url is set. 
-  *  class @param this.disableFloodFill must equal true before called because 
-  *    maskImageData is being updated.
-  *  class @param this.disableSubmit must equal true before called because 
-  *    maskUrl is being updated in drawMask(). 
+  *  class @param this.disableFloodFill must equal true before pixels updated because
```
dtjanaka, line 165: nit "because so"

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -302,32 +300,84 @@ export class EditorComponent implements OnInit {
...
+  updateTolerance(value: number) {
+    this.tolerance = value;
+    console.log('new tolerance: ' + value);
+  }
+
+
+ /** 
+  *  Retrieves new tint value from child component toolbar and draws mask with new tint.
+  *  If the tint value cannot be larger than 1 and if the value inputed is greater than 1, 
+  *    then the tint defaults to 1.
```
robotsaurus, line 320: nit: 'If' isn't needed at the beginning of the sentence.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -302,32 +300,84 @@ export class EditorComponent implements OnInit {
...
+  *        Would decrease lag.
+  */
+  drawPixel(pixel: number) {
+    //Commented out would be the TODO but the pixels don't line up 1:1 so there would be a visual difference in what the 
+    /// user thinks they're clicking on and what they actully paint on the mask.
+    // let imageData = this.ctx.getImageData(0,0,this.canvas.nativeElement.width, this.canvas.nativeElement.height);
+    // imageData.data[pixel] = 255;
+    // imageData.data[pixel + 2] = 255;
+    // imageData.data[pixel + 3] = 255;
+    // this.ctx.putImageData(imageData, 0,0);
```
robotsaurus, line 376: Can this commented out code be removed?  The following seems to do the same thing and it reuses the drawMask function instead of doing it on its own.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -42,12 +43,12 @@ export class EditorComponent implements OnInit {
   scaleFactor: number;
   originalImageData: ImageData;
   tolerance: number;
+  tint:number;
```
robotsaurus, line 46: this is used to set the globalAlpha so I think it is better to call this alpha or maskAlpha instead of introducing a new word tint because I immediately wondered what it was since I didn't expect the word.

src/main/ngapp/accio/src/app/editor/mask.directive.ts
```diff
@@ -10,41 +11,132 @@ export class MaskDirective {
...
   @Input() scale: number;
   @Input() tolerance: number;
-  @Input() disable: boolean;
+  @Input() disableFloodFill: boolean;
+  @Input() tool: MaskTool;
 
   @Output() newMaskEvent = new EventEmitter<Set<number>>();
+  @Output() newPaintEvent = new EventEmitter<number>();
+
+  //  Set containing pixels converted to theit red index in ImageData. Used for paint and scribble
```
robotsaurus, line 20: nit: theit to their

src/main/ngapp/accio/src/app/editor/mask.directive.ts
```diff
@@ -10,41 +11,132 @@ export class MaskDirective {
...
   /** 
-   *  Listens for user interaction on 'appMask' - the  canvas visual to the user.
-   *  When clicked, records the x and y offset, retrieves the tolerance and calls the 
-   *  floodfill algorithm. Directive is a child of editor, so when the new set mask is 
-   *  created, the pixels are emitted to the editor.
-   */ 
-  @HostListener('click', ['$event'])
-  onClick(e: MouseEvent) {
-    if(!this.disable) {
+   *  Listens for mouse movment over appMask, executes if user's mouse is clicked.
```
robotsaurus, line 66: nit: movment to movement

src/main/ngapp/accio/src/app/editor/mask.directive.ts
```diff
@@ -10,41 +11,132 @@ export class MaskDirective {
...
+      console.log('drawing pixel mousemove');
+      //  Fire event to draw pixel
+      this.newPaintEvent.emit(pixel);
+    }
+  }
+
+ /** 
+
+  *  Sends class @param paintPixels of pixels user painted to magic-wand.service
+  */
```
robotsaurus, line 96: these comments seem out of place

src/main/ngapp/accio/src/app/editor/mask.directive.ts
```diff
@@ -10,41 +11,132 @@ export class MaskDirective {
...
+    else if (this.tool == MaskTool.magicWand && !this.scribbleFill) {
       console.log('tolerace in mask.dr ' + this.tolerance);
       
       //  Returns an array indices of each pixel in the mask.
       const maskPixels = this.magicWandService.floodfill(
         this.originalImageData, 
-        Math.floor(xCoord / this.scale), 
-        Math.floor(yCoord / this.scale), 
+        Math.floor(this.xCoord / this.scale), 
+        Math.floor(this.yCoord / this.scale),
```
robotsaurus, line 136: Nothing that you need to change but just mentioning that you might consider creating a function that will take the mouse x,y coordinates and converts it to pixels. Right now this is a math operation because all you need to do is divide by the scale.  However, once you have pan and zoom functionality it will become more complicated.  So when you add that feature all you would have to do it modify that in one place and all of your code is automatically updated.  Contrast that with you having to search over your code for Math.floor and figure out if you need to replace it.

Again you don't need to change it in this PR.  I'm just mentioning it for future guidance.

src/main/ngapp/accio/src/app/toolbar/toolbar.component.ts
```diff
@@ -9,18 +9,21 @@ import { Output, EventEmitter } from '@angular/core';
...
   @Output() clearMaskEvent = new EventEmitter<void>();
-  @Output() newToleranceEvent = new EventEmitter<number>();
   @Output() newMaskToolEvent = new EventEmitter<string>();
   @Output() invertMaskEvent = new EventEmitter<void>();
+  @Output() newToleranceEvent = new EventEmitter<number>();
+  @Output() newTintEvent = new EventEmitter<number>();
 
-  value: number;
+  toleranceValue: number;
+  tintValue: number;
```
robotsaurus, line 18: alphaValue seems like a more consistent name.

src/main/ngapp/accio/src/app/editor/mask.directive.ts
shmcaffrey: noted! I can do that for the next pr.

shmcaffrey:
PTAl

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -302,32 +300,70 @@ export class EditorComponent implements OnInit {
...
+    console.log('new tolerance: ' + value);
+  }
+
+
+ /** 
+  *  Retrieves new alpha value from child component toolbar and draws mask with new alpha.
+  *  The alpha value cannot be larger than 1 or less than 0, so the value is adjusted to fit in range.
+  */
+  updateMaskAlpha(value: number) {
+    this.maskAlpha = (Math.abs(value) % 1);
```
robotsaurus, line 322: I don't think this is exactly what you want here as 1.2 % 1 will be 0.2  I think instead you want to clamp so:

min(max(value, 0.0), 1.0)

shmcaffrey:
ptal

src/main/ngapp/accio/src/app/editor/MaskToolEnum.ts
```diff
@@ -0,0 +1,5 @@
+export enum MaskTool {
+  magicWand = 'MAGIC-WAND',
```
snazzorama, line 2: The enum values (`magicWand`, etc.) should be in upper case with underscores (e.g. `MAGIC_WAND`).

https://google.github.io/styleguide/jsguide.html#naming-enum-names

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -302,32 +300,69 @@ export class EditorComponent implements OnInit {
...
     }
+    console.log('switched tool to ' + this.maskTool);
+    //  Always redraw mask/image when switching between features because of MaskOnly tool.
+    this.drawMask();
+  }
+
+ /**
+  *  Adds pixel user painted to mask.
+  *  TODO: Possibly change the implementation so the pixel drawn is
+  *        replecated on the screen ASAP, and then once the user finishes
```
snazzorama, line 358: replecated -> replicated

src/main/ngapp/accio/src/app/editor/mask.directive.ts
```diff
@@ -10,41 +11,127 @@ export class MaskDirective {
...
 
   //  Create direct reference of canvas on editor.html
   constructor(
     private canvas: ElementRef<HTMLCanvasElement>, 
     private magicWandService: MagicWandService,
   ) { }
 
+/**
+ * Listens for user mousedown on 'appMask'. When user mouse goes down
+ * It initialized the set and paints the single pixel.
```
snazzorama, line 38: This second sentence doesn't quite make sense. I'd suggest: "Initializes the \`paintPixels\' set and paints a single pixel when the user presses the mouse."

src/main/ngapp/accio/src/app/editor/editor.component.ts
robotsaurus: Did you forget to push these new changes?  Or am I thinking about this wrong?

snazzorama:
> There's the one comment on clamping that still needs to be resolved but otherwise this is good.

I see it as `this.maskAlpha = Math.min(Math.max(value, 0.0), 1.0);`... maybe you're looking at an old version?

robotsaurus:
> > There's the one comment on clamping that still needs to be resolved but otherwise this is good.
> 
> I see it as `this.maskAlpha = Math.min(Math.max(value, 0.0), 1.0);`... maybe you're looking at an old version?

Huh! For me right above comment

> Did you forget to push these new changes? Or am I thinking about this wrong?

It shows `this.maskAlpha = (Math.abs(value) % 1);`

What does it show for you?

If I go into the code, then it does indeed show `this.maskAlpha = Math.min(Math.max(value, 0.0), 1.0);`.  So it must have been a loading issue.  I'll check the url in the future to ensure nothing funky is going on.

robotsaurus:
> > > There's the one comment on clamping that still needs to be resolved but otherwise this is good.
> > 
> > 
> > I see it as `this.maskAlpha = Math.min(Math.max(value, 0.0), 1.0);`... maybe you're looking at an old version?
> 
> Huh! For me right above comment
> 
> > Did you forget to push these new changes? Or am I thinking about this wrong?
> 
> It shows `this.maskAlpha = (Math.abs(value) % 1);`
> 
> What does it show for you?
> 
> If I go into the code, then it does indeed show `this.maskAlpha = Math.min(Math.max(value, 0.0), 1.0);`. So it must have been a loading issue. I'll check the url in the future to ensure nothing funky is going on.

Ok, I looked at it further github does say "outdated" on the upper right.  So that is indeed how it happened.

Sorry Shaunna.

PR #35, opened 2020-08-17
Add blobs post tests, format expected outputs for blobs get tests, and add child deletion.

dtjanaka:


src/main/java/com/google/sps/DataUtils.java
```diff
@@ -121,5 +122,37 @@ public static Entity getProjectEntity(String projId, String userEmail,
...
+      deleteImageAndChildren(child.getKey());
+    }
+    datastore.delete(projectKey);
+  }
+
+  /**
+   * Removes an image and all of its children from the database.
+   * @param     {Key}       imgKey  
+   */
+  public static void deleteImageAndChildren(Key imgKey) {
```
shmcaffrey, line 145: Children being masks, right?

dtjanaka: yes

src/main/java/com/google/sps/BlobUtils.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * Provides utilities for the blobs servlet.
+ */
+public final class BlobUtils {
+
+  /**
+   * Checks if the file uploaded to Blobstore is valid.
```
snazzorama, line 32: Small suggestion: state what's going to happen if the file is not valid.

PR #36, opened 2020-08-18
Initial PR for scribbleFloodfill algorithm

charmychancey:
- Change tolerance logic to compare tolerance value based on euclidean distance instead of comparing each individual R, G, 
  and B value.
- Update appropriate comments
- Implement an initial scribbleFloodfill() (waiting for brush UI to be merged to master so that scribbleFloodfill() can be better 
  tested and optimized).
    - Add helper method pixelIndexToXYCoord()
        - Converts pixel index {number} into an [x, y] coodinate {Array<number> [x, y]}
    - Implement helper method getIsScribbleMask()
        - Adds the pixel to the mask as long as it satisfies the color tolerance for at least 1 pixel in the set of reference pixels.

### DIFF
--- a/src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
+++ b/src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
@@ -73,16 +73,13 @@ export class MagicWandService {
     const curX: number = pixelCoord[0];
     const curY: number = pixelCoord[1];
 
-    // Gets array of the color attributes of current pixel.
+    // Gets [R, G, B] of current pixel.
     const curPixel: Array<number> = this.dataArrayToRgba(imgData, curX, curY);
 
-    // Color attributes of pixel (R,G, and B) must be within tolerance level.
-    for (let i = 0; i < 3; i++) {
-      const upperTolerance: boolean = curPixel[i] > originalPixel[i] + tolerance;
-      const lowerTolerance: boolean = curPixel[i] < originalPixel[i] - tolerance;
-      if (upperTolerance || lowerTolerance) {
-        return false;
-      }
+    const colorDifference = this.rgbEuclideanDist(originalPixel, curPixel);
+
+    if (colorDifference > tolerance) {
+      return false;
     }
 
     return true;
@@ -92,7 +89,7 @@ export class MagicWandService {
    * @param {Array<number> [R, G, B]} basisColor and 
    * @param {Array<number> [R, G, B]} secondColor
    */
-  RgbEuclideanDist(basisColor: Array<number>, secondColor: Array<number>)
+  rgbEuclideanDist(basisColor: Array<number>, secondColor: Array<number>)
       : number {
     if (basisColor.length != secondColor.length) {
       throw new Error(
--- a/src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
+++ b/src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
@@ -252,7 +252,7 @@ export class MagicWandService {
         }
         // Visits the pixel and check if it should be part of the mask.
         visited.add(this.coordToDataArrayIndex(x, y, imgData.width));
-        if (this.getIsScribbleMask(originalPixel, imgData, neighborPixel, tolerance)) {
+        if (this.getIsScribbleMask(originalPixel, imgData, neighborPixel, tolerance, scribbles)) {
           visit.push(neighborPixel);
         }
       }
@@ -284,12 +284,22 @@ export class MagicWandService {
     return true;
   }
 
-  filterScribbles(scribbles: Set<number>, tolerance: number): Set<number> {
+  /**@returns {Set<number>} filtered set of pixels by removing indices 
+   * of pixels whose color are too similar to the original pixel. 
+   * Judgement of 'too similar' is decided by a factor of 
+   * @param {number} tolerance and color value itself is retrieved from
+   * @param {ImageData} imgData
+  */
+  filterScribbles(scribbles: Set<number>, originalPixel: Array<number>, 
+      imgData: ImageData, tolerance: number): Set<number> {
     // TODO
     let result: Set<number> = new Set();
     for (let pixelCoord of scribbles) {
-      pixelColor = RgbEuclideanDist();
-      if (/* filter condition */) {
+      const x = (pixelCoord / 4) % imgData.width;
+      const y = (pixelCoord / 4) / imgData.width;
+      const curPixel = this.dataArrayToRgba(imgData, x, y);
+      const colorDifference = this.rgbEuclideanDist(originalPixel, curPixel);
+      if (colorDifference > tolerance / 2) {
         result.add(pixelCoord);
       }
     }
--- a/src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
+++ b/src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
@@ -77,8 +77,10 @@ export class MagicWandService {
     const curPixel: Array<number> = this.dataArrayToRgb(imgData, curX, curY);
 
     const colorDifference = this.rgbEuclideanDist(originalPixel, curPixel);
+    // Work with tolerance logic in squared space for Euclidean distance.
+    const squaredTolerance = tolerance * tolerance;
 
-    if (colorDifference > tolerance) {
+    if (colorDifference > squaredTolerance) {
       return false;
     }
 
@@ -106,7 +108,10 @@ export class MagicWandService {
       distance += Math.pow((basisColor[i] - secondColor[i]), 2);
     }
 
-    return Math.sqrt(distance);
+    // Does not sqrt distance to complete eucidean dist formula b/c sqrt is
+    // an expense operation. Instead, can compare against tolerance in the 
+    // squared space.
+    return distance;
   } 
 
   // Checks if @pixelCoord is in bounds and makes sure it's not a repeat coord.
@@ -271,6 +276,8 @@ export class MagicWandService {
 
     // Gets array of color attributes of current pixel.
     const curPixel: Array<number> = this.dataArrayToRgb(imgData, curX, curY);
+    // Work with tolerance logic in squared space for Euclidean distance.
+    const squaredTolerance = tolerance * tolerance;
 
     for (let pixelIndex of scribbles) {
       const refPixelCoord: Array<number> = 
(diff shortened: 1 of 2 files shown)


src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -175,4 +201,130 @@ export class MagicWandService {
...
+   */
+  getIsScribbleMask(scribbles: Set<number>, imgData: ImageData,
+      pixelCoord: Array<number>, tolerance: number): boolean {
+    const curX: number = pixelCoord[0];
+    const curY: number = pixelCoord[1];
+
+    // Gets array of color attributes of current pixel.
+    const curPixel: Array<number> = this.dataArrayToRgba(imgData, curX, curY);
+
+    for (let pixelIndex of scribbles) {
```
shmcaffrey, line 275: Would scribbles be better passed in as a set of (x, y) pairs? That way you don't have to translate them back.

shmcaffrey:
lgtm but I think what we were talking about yesterday in the pod meeting will be important to consider. How if you scrub out, say, a blue-green color because the tolerance is within the euclidean color distance of the blue-green color to the original pixel that's say, blue, you might miss some of the values in the mask making process. I think it's just something we'll have to wait to see when it's connected with the UI. It could be worth scrubbing out the pixels that aren't within the initial tolerance because then you know you'd reach all of the similar colors in one swoop instead of some similar colors and then a gap, and then more similar colors. But then you'd have a lot of overlap as well, so I see your reasoning.

charmychancey:
> lgtm but I think what we were talking about yesterday in the pod meeting will be important to consider. How if you scrub out, say, a blue-green color because the tolerance is within the euclidean color distance of the blue-green color to the original pixel that's say, blue, you might miss some of the values in the mask making process. I think it's just something we'll have to wait to see when it's connected with the UI. It could be worth scrubbing out the pixels that aren't within the initial tolerance because then you know you'd reach all of the similar colors in one swoop instead of some similar colors and then a gap, and then more similar colors. But then you'd have a lot of overlap as well, so I see your reasoning.

Right, and I agree that this will be an optimization issue that we can tinker with after the scribbleFloodfill is integrated into the editor component.
And I think in the end, some data just has to be scrubbed out, and we just have to find the best approach by tinkering with it on the UI side until we reach a point that feels intuitive for the user. For example, the paper that I referred to for the idea of scribbleFloodfill had their own method of "scrubbing" out data by condensing "circles" with radius of (for example) 50px into a single datapoint with an averaged color of the pixels within that circle. 
My idea is that this approach to behind-the-scenes scrubbing is probably less intuitive for the user, and it also offers them less control. I wanted to try and let the user have more control over exactly what pixel color is added to the reference-set for tolerance-checks, because, if we went with a circle-color-averaging approach, then the user would be losing control over what the "true" color of the pixel that they were trying to select was. 
When I considered the practical application of scribbleFloodfill, I also thought of it's primary usefulness to be in situations where the user would want to the floodfill and expand to pixels with colors that were obviously outside of their current tolerance threshold, and thus significantly beyond the range color-data that my approach would be scrubbing out. The reason being is, if they wanted colors extremely close to the initial pixel, it would make more sense to just allow the user to adjust their tolerance input with a basic floodfill() command. Furthermore, the problem of losing color-data for tolerance-references can be a flexible experience for the user, because they can adjust the tolerance to smaller distances, which would decrease the amount of colors excluded from the color-data tolerance-referenced set.
We'll definitely talk more about this once we start testing it within the editor itself.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
charmychancey: Nope, I don't mind; it's all good since you already have them in Set<number> format on your side, I was able to simply handle it with a helper function on my end that converts DataArray indexing into [x, y] format.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -7,95 +7,116 @@ import { SetOperator } from './set-operator';
...
-     * Store already-visited pixels in "visited" as index formatted numbers
-     * (as opposed to coord format; for Set funcs).
-     */
+    // Stores a queue of coords for pixels that we need to visit in "visit".
     const visit: Array<Array<number>> = new Array();
+    // Stores already-visited pixels in "visited" as index formatted numbers
+    // (as opposed to coord format; for Set funcs).
     const visited: Set<number> = new Set();
-    // Use a set for mask; mainly do iter and set operations on masks
+    // Uses a set for mask; mainly do iter and set operations on masks...
```
robotsaurus, line 21: nit: three periods

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -7,95 +7,116 @@ import { SetOperator } from './set-operator';
...
+    // Loops until no more adjacent pixels within tolerance level.
     while (visit.length !== 0) {
       const coord: Array<number> = visit.pop();
-      // Unpack coord
+      // Unpacks coord
       const x: number = coord[0];
       const y: number = coord[1];
 
-      // Operational part of while-loop
+      // Operational part of while-loop...
```
robotsaurus, line 41: nit: three periods. Should only be one as you want full sentences.  Here and below.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -7,95 +7,116 @@ import { SetOperator } from './set-operator';
...
+          'Arguments must be an array of [R, G, B] (length == 3)...');
+    }
+
+    let distance = 0;
+
+    for (let i = 0; i < basisColor.length; i++) {
+      distance += Math.pow((basisColor[i] - secondColor[i]), 2);
+    }
+
+    return Math.sqrt(distance);
```
robotsaurus, line 109: sqrt is an expensive operation so I would suggest just working in the squared space.  So if you square your tolerance you can accomplish the same thing.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -175,4 +201,130 @@ export class MagicWandService {
 
     return invertedMask;
   }
+
+
+  /* Smarter flood fill tool */
+
```
robotsaurus, line 207: Update comment about what makes it smarter.

robotsaurus: Also I am not sure what scribbles is.  How is that different than xCoord and yCoord?

charmychancey: Scribbles is a set of pixels, whose colors are used to expand the reference pixels that the tolerance logic uses to determine if a pixel should be part of the mask or not. The range of our tolerance is essentially being expanded to include colors within the scribbles set.
I've updated this to be more clear in the in-line function documentation.

charmychancey:
@robotsaurus Ptal

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -175,4 +159,184 @@ export class MagicWandService {
...
+            continue;
+          }
+          // Visits the pixel and check if it should be part of the mask.
+          visited.add(this.coordToDataArrayIndex(x, y, imgData.width));
+          if (this.getIsScribbleMask(scribbles, imgData, neighborPixel, tolerance)) {
+            visit.push(neighborPixel);
+          }
+        }
+      }  // End of while loop.
+    }
```
robotsaurus, line 338: We are getting close.  If you notice the if/else code is exactly the same code except for:

`if (this.getIsMask(originalPixel, imgData, neighborPixel, tolerance))`

and 

`if (this.getIsScribbleMask(scribbles, imgData, neighborPixel, tolerance))`

So you could do two things:

Remove outer if/else and add it to lines 300.  So it would be:

```
if (scribbles.size === 0) {
  if (this.getIsMask(originalPixel, imgData, neighborPixel, tolerance)) {
    visit.push(neighborPixel);
   }
} else {
  if (this.getIsScribbleMask(scribbles, imgData, neighborPixel, tolerance)) {
    visit.push(neighborPixel);
  }
}
```

I think you might be able to do even better.  

One option is to create a helper function:

```
getIsMask(originalPixel, scribbles,  imgData, neighborPixel, tolerance) {
  if (scribbles.size === 0) {
    if (this.getIsMask(originalPixel, imgData, neighborPixel, tolerance)) {
      visit.push(neighborPixel);
     }
  } else {
    if (this.getIsScribbleMask(scribbles, imgData, neighborPixel, tolerance)) {
      visit.push(neighborPixel);
    }
  }
}
```

This is better because then the floodfill code is easier to read, but can we do better?

If I understand it correctly I think we can get the same functionality out of getIsMask by using getIsScribbleMask. If we call `this.getIsScribbleMask(new Set<number>(originalPixel), imgData, neighborPixel, tolerance)` then it will return the same thing.

So I would recommend that you change floodFill to:

```
floodfill(imgData: ImageData, xCoord: number, yCoord: number,
      tolerance: number): Set<number> {
    scribbles = new Set<number>(this.pixelIndexToXYCoordToPixelIndex(xCoord, yCoord))
    return this.doFloodfill(imgData, xCoord, yCoord, tolerance, scribbles);
  }
```

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -175,4 +159,184 @@ export class MagicWandService {
...
+  pixelIndexToXYCoord(pixelIndex: number, width: number): Array<number> {
+    return [((pixelIndex / 4) % width), ((pixelIndex / 4) / width)];
+  }
+
+  /**Does the general floodfill algorithm, and switches to 
+   * scribbleFloodfill() if 
+   * @param {Set<number>} scribbles is provided as a non-empty set.
+   */
+  doFloodfill(imgData: ImageData, xCoord: number, yCoord: number,
+      tolerance: number, scribbles: Set<number>): Set<number> {
```
robotsaurus, line 254: You don't need to do this in this PR, but something to consider for future clean ups.  You have a mixed representation of pixel coordinates.  They are either (x,y) or pixel indices.  You could consider picking one and using it throughout all headers.  By making this consistent it makes it easier to work with and to know what to pass in.

charmychancey: Thanks! I'll note this down.

PR #37, opened 2020-08-19
Editor

shmcaffrey:
Added the ability to move between images on the editor page. Right now if the user refreshed the page on editor, the array holding the images is lost so they have to go back to the gallery page. I want to fix the canvas and add more features but this is something to come back to.


src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.html
```diff
@@ -65,16 +65,16 @@ <h1>Search by:</h1>
 <!--Filter Bar-->
 
 <!--Fetched Images-->
-<div class="gallery">
-  <div *ngFor="let image of imageArray" class="gallery-item">
-    <a [routerLink]="['/editor', projectId, image['name'], image['url']]">
+<div class="gallery" *ngIf="displayImages">
```
charmychancey, line 68: What is the ngIf for? Does it actually change the behavior that the user sees, or is it the same without it? Wondering because I also had an ngIf here originally for the purposes at the time of just fetching images for the img gallery, but I removed it because there didn't seem to be any problems without it.

src/main/ngapp/accio/src/app/img-gallery/img-gallery.component.html
```diff
@@ -65,16 +65,16 @@ <h1>Search by:</h1>
...
+  <div *ngFor="let image of imageArray; index as i" class="gallery-item">
+    <a [routerLink]="['/editor', projectId, image['name'], image['url'], i]">
       <img class="gallery-img" [src]="image['url']">
       <p>{{ image['name'] }}</p>
     </a>
   </div>
 </div>
 
-<div class="gallery">
+<div class="gallery" *ngIf="displayImages">
```
charmychancey, line 77: Same q here

charmychancey: other than that, lgtm

PR #53, opened 2020-08-24
Floodfill kd tree

charmychancey:
Implement floodfill algos using kd-tree:
- Add external library for kd tree: 
  - MIT License <http://www.opensource.org/licenses/mit-license.php>
- Remove/change old functions
  - rgbEuclideanDist() => changed
    - Takes in Color objects as opposed to pixel coordinates now.
  - getIsScribbleMask() => removed
  - getIsMask() => removed
  - filterScribbles() => removed
- Add function scribblesToColors()
  - Convert Array of pixels to Array of Colors.
- Add colorDistance as an alternate to rgbEuclideanDist()
  - http://www.compuphase.com/cmetric.htm
  - Taken from an external study conducted on colour metrics
  - Part of MIT License <http://www.opensource.org/licenses/mit-license.php> module code
- Adapt doFloodfill() method for kd-tree implementation


src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -298,12 +212,54 @@ export class MagicWandService {
...
+   */
+  rgbEuclideanDist(colorA: Color, colorB: Color): number {
+
+    let distance = 0;
+
+    for (let attribute in colorA) {
+      distance += Math.pow((colorA[attribute] - colorB[attribute]), 2);
+    }
+
+    // Does not sqrt distance to complete eucidean dist formula b/c sqrt is
```
dtjanaka, line 242: nit: Euclidean

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -298,12 +212,54 @@ export class MagicWandService {
...
+    // squared space.
+    return distance;
+  }
+
+  /**@license MIT License <http://www.opensource.org/licenses/mit-license.php>
+   * Pretty good color distance from
+   * http://www.compuphase.com/cmetric.htm
+   */
+  colorDistance(a: Color, b: Color): number {
+    var dr = a.red - b.red;
```
dtjanaka, line 253: not sure if this applies for typescript/angular, but https://google.github.io/styleguide/jsguide.html#features-use-const-and-let mentions using const or let over var.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -298,12 +212,54 @@ export class MagicWandService {
         }
         // Visits the pixel and check if it should be part of the mask.
         visited.add(this.coordToDataArrayIndex(x, y, imgData.width));
-        if (this.getIsScribbleMask(scribbles, imgData, neighborPixel, tolerance)) {
+        // if (this.getIsScribbleMask(scribbles, imgData, neighborPixel, tolerance)) {
+        //   visit.push(neighborPixel);
+        // }
+        if (tree.nearest(
+            this.scribblesToColors(new Set([
+            this.coordToDataArrayIndex(x, y, imgData.width)]), imgData)[0],
```
snazzorama, line 220: It's pretty confusing to use the scribblesToColors function this way. I would suggest having a different function that turns a pixel index into a Color object that you can call here. scribblesToColors can also call that function.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -298,12 +212,54 @@ export class MagicWandService {
         }
         // Visits the pixel and check if it should be part of the mask.
         visited.add(this.coordToDataArrayIndex(x, y, imgData.width));
-        if (this.getIsScribbleMask(scribbles, imgData, neighborPixel, tolerance)) {
+        // if (this.getIsScribbleMask(scribbles, imgData, neighborPixel, tolerance)) {
+        //   visit.push(neighborPixel);
+        // }
```
snazzorama, line 217: If this code is unused, you should just remove it. If you need to refer to it later, you can always look in the git history.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -298,12 +212,54 @@ export class MagicWandService {
...
+   * @param {Array<number> [R, G, B]} basisColor and 
+   * @param {Array<number> [R, G, B]} secondColor
+   */
+  rgbEuclideanDist(colorA: Color, colorB: Color): number {
+
+    let distance = 0;
+
+    for (let attribute in colorA) {
+      distance += Math.pow((colorA[attribute] - colorB[attribute]), 2);
+    }
```
snazzorama, line 240: Since this is going to be called a lot, we might want to optimize a little. Also, I find the loop a little unclear -- and it could do surprising things if there were extra properties. I'd suggest calculating dr, dg, and db like you do below and returning dr*dr + dg*dg + db*db.

src/main/ngapp/accio/src/app/editor/kdTree.js
```diff
@@ -0,0 +1,465 @@
+/**
```
robotsaurus, line 1: You don't need to break up this PR.  It is better to add third party dependencies in a standalone PR. This separates packages from code.

Additionally, the best thing to do is to commit the unmodified library.  Then submit an additional PR if you need to make any modifications to the library to make it compile/work in your repository.  Your future self will thank you when you need to the exact same task again.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -255,16 +165,16 @@ export class MagicWandService {
    */
   doFloodfill(imgData: ImageData, xCoord: number, yCoord: number,
       tolerance: number, scribbles: Set<number>): Set<number> {
+    // Creates an array of color objects from the indices in 'scribbles'.
+    // Color object: {red: number, green: number, blue: number}
+    const colors: Array<Color> = this.scribblesToColors(scribbles, imgData);
```
robotsaurus, line 170: I think you should use a more descriptive variable name here.  Something like: scribblesColors or thresholdColors etc.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
charmychancey: Thanks for pointing this out! I then realized that I already have an existing function that does this, I just need to modify it a bit.

src/main/ngapp/accio/src/app/editor/kdTree.js
charmychancey: Sounds good, Thanks!

charmychancey:
@robotsaurus @snazzorama ptal

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -1,309 +1,270 @@
...
+       if (!this.getIsValid(imgData.width, imgData.height, x, y, visited)) {
+         continue;
+       }
+       // Visits the pixel and check if it should be part of the mask.
+       visited.add(this.coordToDataArrayIndex(x, y, imgData.width));
+
+       /* Ref: kdTree.nearest(point, count, maxDistance) */
+       if (tree.nearest(
+         this.dataArrayToRgb(imgData, x, y),
+         1, tolerance * tolerance).length > 0) {
```
robotsaurus, line 228: compute the tolerance square once.

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -1,309 +1,270 @@
...
+   // Converts [x,y] format coord to 1-D equivalent of
+   // imgData.data (DataArray).
+   const indexAsDataArray: number =
+     this.coordToDataArrayIndex(xCoord, yCoord, imgData.width);
+   visited.add(indexAsDataArray);
+
+   const tree = new KdTree.kdTree(
+     thresholdColors, this.rgbEuclideanDist, ["red", "green", "blue"]);
+
+
```
robotsaurus, line 198: nit: double newline

PR #54, opened 2020-08-25
Implement cursor-resizing to complement brush-resizing

charmychancey:
- Add additional canvas on top of current scaledCanvas
  - This new canvas(cursorCanvas) paints a new 'cursor' each time the mouse moves
- Disable pointer-events for cursorCanvas in .css file
  - cursorCanvas gets the data for mouse position that's needed to paint the
    cursor from scaledCanvas.
- Add helper functions setCursorPosition() and updateCursor()
  - Use to help paint and update 'cursor'
- Add code to initCanvas() to initiate new cursorCanvas


src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -84,6 +89,13 @@ export class EditorComponent implements OnInit {
   scaledCanvas: ElementRef<HTMLCanvasElement>; 
   private scaledCtx: CanvasRenderingContext2D;
 
+  // Overlays custom "cursor" that changes based on brush size.
+  @ViewChild('cursorCanvas', { static: true })
+  cursorCanvas: ElementRef<HTMLCanvasElement>;
+  private cursorCtx: CanvasRenderingContext2D;
+  cursorTarget;
```
shmcaffrey, line 96: Add types? Should just be ElementRef\<HTMLCanvasElement\>

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -166,37 +178,72 @@ export class EditorComponent implements OnInit {
...
     this.stageWidth = imgWidth * this.scaleFactor;
     this.stageHeight = imgHeight * this.scaleFactor;
 
-    //   Draws image non scaled on full canvas
+    // Canvas to paint cursor-overlay of brush size.
+    this.cursorCanvas.nativeElement.width = imgWidth * this.scaleFactor;
+    this.cursorCanvas.nativeElement.height = imgHeight * this.scaleFactor;
+
+    // Updates the drawn 'cursor' when the user's mouse moves.
+    this.maskTarget = document.querySelector('#mask-layer');
```
shmcaffrey, line 206: Is there a reason you're not using cursorCanvas/maskCanvas? as in, this.cursorCtx = this.cursorCanvas.getContext('2d') and maskcanvas.addeventlistener?

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -166,37 +178,72 @@ export class EditorComponent implements OnInit {
...
+    this.cursorTarget = document.querySelector('#cursor-layer');
+    this.cursorCtx = this.cursorTarget.getContext('2d');
+    this.maskTarget.addEventListener('mousemove', (e) => {
```
shmcaffrey, line 209: I think the mouse move and mouse out events should technically be handled in the mask.directive because they are already in use there, that would mean passing the ctx to the directive as an input, which I believe can be done.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -84,6 +89,13 @@ export class EditorComponent implements OnInit {
   scaledCanvas: ElementRef<HTMLCanvasElement>; 
   private scaledCtx: CanvasRenderingContext2D;
 
+  // Overlays custom "cursor" that changes based on brush size.
+  @ViewChild('cursorCanvas', { static: true })
+  cursorCanvas: ElementRef<HTMLCanvasElement>;
+  private cursorCtx: CanvasRenderingContext2D;
+  cursorTarget;
+  maskTarget;
```
shmcaffrey, line 97: Are these necessary? You just make them reference the cursor canvas and scaledCanvas right? but the reference already exists in cursorCanvas and scaledCanvas. So I would just use those references and not make a second element ref.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -166,37 +178,72 @@ export class EditorComponent implements OnInit {
...
 
-    //   Draws image non scaled on full canvas
+    // Canvas to paint cursor-overlay of brush size.
+    this.cursorCanvas.nativeElement.width = imgWidth * this.scaleFactor;
+    this.cursorCanvas.nativeElement.height = imgHeight * this.scaleFactor;
+
+    // Updates the drawn 'cursor' when the user's mouse moves.
+    this.maskTarget = document.querySelector('#mask-layer');
+    this.cursorTarget = document.querySelector('#cursor-layer');
+    this.cursorCtx = this.cursorTarget.getContext('2d');
```
shmcaffrey, line 208: Like this can be `this.cursorCtx = this.cursorCanvas.getContext('2d')`

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -166,37 +178,72 @@ export class EditorComponent implements OnInit {
...
+    this.maskTarget.addEventListener('mousemove', (e) => {
+      this.setCursorPosition(e);
+    }, false);
+    this.maskTarget.addEventListener('mouseout', () => {
```
shmcaffrey, line 212: I think the event listeners can just go in the mask.directive under the respective event listeners right? You could Input the cursorCtx and setCursorPosition can be called in mask.directive. I think you can pass ctx's through inputs. This way the user interaction is separate from the component.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -166,37 +178,72 @@ export class EditorComponent implements OnInit {
...
+  // Draws/Redraws 'cursor' at the current position of user's mouse.
+  setCursorPosition(e) {
+    // These are the coordinates used to paint.
+    this.cursorX = e.offsetX;
+    this.cursorY = e.offsetY;
+
+    this.cursorCtx.clearRect(0, 0, this.stageWidth, this.stageHeight);
+
+    this.cursorCtx.beginPath();
+    this.cursorCtx.arc(this.cursorX, this.cursorY,
```
shmcaffrey, line 238: I think if you scale the ctx: 
`this.cursorCtx.scale(this.scaleFactor, this.scaleFactor);`
You can avoid having to set the radius to brushWidth * scaleFactor * .5. because it should automatically scale the radius for you and be the proper size. You may have to play around with this some, I'm not entirely sure.

src/main/ngapp/accio/src/app/editor/editor.component.ts
shmcaffrey: Or you could Output the event and catch it with the setCursorPosition, we can talk about this if you want to!

src/main/ngapp/accio/src/app/editor/editor.component.ts
charmychancey: I believe you did it the angular way, which didn't work with how I implemented it, typescript doesn't like the typing. Thus, I had to make new vars that essentially referenced the same thing in order to use the event listener methods.

src/main/ngapp/accio/src/app/editor/editor.component.ts
charmychancey: Same as above explanation

src/main/ngapp/accio/src/app/editor/editor.component.ts
charmychancey: I think drawing a circle wasn't matching the brush stroke size even on un-scaled images. It was still about twice as big, which is why I needed to scale circle radius by .5

src/main/ngapp/accio/src/app/editor/editor.component.ts
charmychancey: Moved implementation into mask directive

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -166,37 +175,67 @@ export class EditorComponent implements OnInit {
...
     this.imageCanvas.nativeElement.width = imgWidth * this.scaleFactor;
     this.imageCanvas.nativeElement.height = imgHeight * this.scaleFactor;
     this.imageCtx = this.imageCanvas.nativeElement.getContext('2d');
-    
+
     this.stageWidth = imgWidth * this.scaleFactor;
     this.stageHeight = imgHeight * this.scaleFactor;
 
-    //   Draws image non scaled on full canvas
+    // Canvas to paint cursor-overlay of brush size.
```
shmcaffrey, line 198: Can you move this above where stage height is initialized? just for cohesion.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -166,37 +175,67 @@ export class EditorComponent implements OnInit {
...
+  setCursorPosition(e: MouseEvent): void {
+    // These are the coordinates used to paint.
+    this.cursorX = e.offsetX;
+    this.cursorY = e.offsetY;
+
+    this.cursorCtx.clearRect(0, 0, this.stageWidth, this.stageHeight);
+
+    this.cursorCtx.beginPath();
+    this.cursorCtx.arc(this.cursorX, this.cursorY,
+        this.brushWidth * this.scaleFactor * .5, 0, 2 * Math.PI, true);
```
snazzorama, line 226: Does the cursor become invisible if the radius is too small? I wonder if we want to clamp it to be at least 1.

charmychancey: The cursor is still visible at the smallest brush size, which would be a value of 1.

snazzorama: But what if scaleFactor is less than 1? I can definitely make the cursor disappear on the live version of the app.

PR #57, opened 2020-08-27
Merge branch

shmcaffrey:
Pr's from a few branches that were being worked on during user tests.
Incorporates, floodfill using KD tree, style changes, hot keys, toolbar changes, among other things and minor bug fixes.


src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -238,22 +254,34 @@ export class EditorComponent implements OnInit {
...
+      };
       maskImage.src = this.maskUrl;
     }
-  }
 
+    this.allPixels = new Set([
+      ...Array.from(Array(this.originalImageData.data.length / 4).keys()).map(
+        function (x) {
+          return x * 4;
+        }
```
robotsaurus, line 281: It would be useful to add a comment about what this is doing for future reference.

PR #64, opened 2020-08-30
Preview to frontend

charmychancey:
### **IMPORTANT**: _Shortest paths algo fails in the chrome browser to execute on some images. The page hangs and dies before it finishes executing._
  - I don't know if this is a case of the image solely being too big or also about color-distribution in the image. The algo runs in 8 -12 seconds for an image of 1,000,000 pixels, but fails on an image of 960,000 pixels, so it'probably not due to the image size. 
  ### - Possibilities:
  - This might be a problem with the priority queue adding too many neighbors.
    - CURRENT WORKAROUND: Putting a hard limit on the size of the priority queue allows the algo to run successfully in the browser.
    - @dtjanaka @shmcaffrey feel free to play around with adjusting this hardcoded value. It is in the magicwandservice typescript file, in the while loop of the getShortestPaths() method 

### 8|28|2020(approx)
Implement preview floodfill algo for user-interaction
- Revamp tolerance slider so that it updates real-time as the user is in the process of sliding it.
- Update **PreviewMask class** to handle providing more data for editor.component.
  - **masksAtTolerance()** => **changeMaskBy()**
    - Make void
      - no longer converts the present mask into a new set each time; caller does not necessarily want the set after every call to this function
  - Make **new getter methods** for:
    - getting the present mask as it is (an array)
    - getting the present mask as a set
    - getting a boolean to check if the PreviewMask object has been initialized properly
      - Used by caller to check if they should perform a preview-floodfills or normal floodfill sequence
  - Add **resetMask()**
    - Allows caller to reset this.presentMask to an empty array and also reset this object's tolerance-index-pointer
- Changes to mask action file:
  - Update mask action class to handle passing PreviewMask as data to the editor component
- Changes to mask directive:
  - No longer emitts basic floodfill actions; replaced with preview-floodfill
    - Might want to re-enable basic floodfill later, because it is a lot faster to execute that algo. vs preview-floodfill algo.
    - Still emits scribble floodfill action
    - Still capable of floodfill-subtraction action

### 8|29|2020(approx)
- Changes to app module file:
  - Add import and declaration for snackbar module
- Add new canvas '**previewCanvas**' to editor component
  - Draw the preview separately from the scaledCanvas so that they user can see the new previews of floodfills at different tolerance levels
  - **updatePreview()** and **drawPreview()** methods are used to redraw the previewed-mask at the current user-inputted tolerance each time that the user changes the tolerance (using either the slider or number input next to the slider)
    - updatePreview() triggers the preview-snackbar to pop-up
      - Instruct the user to click 'confirm' on the snackbar if they want to commit the previewed-floodfill
        - Clicking the 'confirm' action on the snackbar triggers **endPreview()**
          - Commit the previewed floodfill at the current tolerance-input into the maskController aka undo/redo logic-flow
      - Block the user from starting another action (point-events stop working on the canvas area) until they press "confirm" on the snackbar

### DIFF
--- a/src/main/ngapp/accio/src/app/editor/editor.component.ts
+++ b/src/main/ngapp/accio/src/app/editor/editor.component.ts
@@ -718,7 +718,7 @@ export class EditorComponent implements OnInit {
       }
       this.drawMask();
       this.disableSubmit = this.disableFloodFill = false;
-      }
+    }
   }
 
   // User must confirm the preview and commit it to the mask before 
--- a/src/main/ngapp/accio/src/app/editor/editor.component.ts
+++ b/src/main/ngapp/accio/src/app/editor/editor.component.ts
@@ -784,7 +784,7 @@ export class EditorComponent implements OnInit {
       this.drawMask();
       this.disableSubmit = this.disableFloodFill = false;
 
-      // Unpauses other canvas-editting sequences from registering
+      // Unpauses other canvas-editing sequences from registering
       // on the canvas. 
       const previewLayer = document.getElementById('preview-layer');
       previewLayer.style.pointerEvents = 'none';
--- a/src/main/ngapp/accio/src/app/editor/editor.component.ts
+++ b/src/main/ngapp/accio/src/app/editor/editor.component.ts
@@ -762,8 +762,6 @@ export class EditorComponent implements OnInit {
       this.continuePreview = false;
       this.clearPreviewCanvas();
 
-      let alphaValue = this.maskTool == MaskTool.MAGIC_WAND_ADD ? 255 : 0;
-
       for (let pixel of this.previewMaster.getMaskAsArray()) {
         this.maskImageData.data[pixel] = 255;
         this.maskImageData.data[pixel + 2] = 255;
--- a/src/main/ngapp/accio/src/app/editor/editor.component.ts
+++ b/src/main/ngapp/accio/src/app/editor/editor.component.ts
@@ -11,6 +11,7 @@ import { Coordinate } from './Coordinate';
 import * as $ from 'jquery';
 import { MaskAction, Action, Tool } from './mask-action';
 import { MaskControllerService } from './mask-controller.service';
+import { PreviewMask } from './magic-wand.service';
 
 @Component({
   selector: 'app-editor',
@@ -99,6 +100,14 @@ export class EditorComponent implements OnInit {
   @ViewChild('scaledCanvas', { static: true })
   scaledCanvas: ElementRef<HTMLCanvasElement>;
   private scaledCtx: CanvasRenderingContext2D;
+
+  // Overlays a layer just for showing the preview of the mask.
+  // Once the preview is committed, this canvas is cleared and 
+  // the chosen preview is painted on the scaledCanvas as well 
+  // as committed to the maskCanavs.
+  @ViewChild('previewCanvas', { static: true })
+  previewCanvas: ElementRef<HTMLCanvasElement>;
+  private previewCtx: CanvasRenderingContext2D;
 
   // Overlays custom "cursor" that changes based on brush size.
   @ViewChild('cursorCanvas', { static: true })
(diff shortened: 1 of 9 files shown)


charmychancey:
@shmcaffrey @dtjanaka after you guys take a look at this, if I'm not back yet, could you also add Brett and M0ike as reviewers?

~Thanks!

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -654,23 +679,140 @@ export class EditorComponent implements OnInit {
...
+      this.previewMaster.resetMask();
+      if (this.curMaskAction.getActionType() == Action.SUBTRACT) {
+        this.maskControllerService.do(this.curMaskAction, this.allPixels);
+      } else {
+        this.maskControllerService.do(this.curMaskAction);
+      }
+      this.drawMask();
+      this.disableSubmit = this.disableFloodFill = false;
+
+      // Unpauses other canvas-editting sequences from registering
```
dtjanaka, line 787: nit: editing

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -100,6 +109,14 @@ export class EditorComponent implements OnInit {
   scaledCanvas: ElementRef<HTMLCanvasElement>;
   private scaledCtx: CanvasRenderingContext2D;
 
+  // Overlays a layer just for showing the preview of the mask.
+  // Once the preview is committed, this canvas is cleared and 
+  // the chosen preview is painted on the scaledCanvas as well 
+  // as committed to the maskCanavs.
```
robotsaurus, line 115: nit: maskCanvas

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -654,23 +679,140 @@ export class EditorComponent implements OnInit {
...
+        this.maskImageData.data[pixel + 3] = alphaValue;
+      }
+      if (maskAction.getActionType() == Action.SUBTRACT) {
+        this.maskControllerService.do(maskAction, this.allPixels);
+      } else {
+        this.maskControllerService.do(maskAction);
+      }
+      this.drawMask();
+      this.disableSubmit = this.disableFloodFill = false;
+      }
```
robotsaurus, line 721: nit: indentation seems wrong here

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -654,23 +679,140 @@ export class EditorComponent implements OnInit {
...
+  }
+  /**Called when a new preview mask sequence and the tolerance input
+   * has been confirmed.
+   */
+  endPreview(): void {
+    if (this.continuePreview) {
+      this.continuePreview = false;
+      this.clearPreviewCanvas();
+
+      let alphaValue = this.maskTool == MaskTool.MAGIC_WAND_ADD ? 255 : 0;
```
robotsaurus, line 765: the alpha value is not set below

src/main/ngapp/accio/src/app/editor/magic-wand.service.ts
```diff
@@ -306,6 +306,12 @@ export class MagicWandService {
 
     // Updates shortest path between neighbor pixel and original node.
     while (toVisit.getSize() !== 0) {
+      // This if statement limits the size of the priority queue for 
+      // reasonable runtime performance in the browser. Removing this 
+      // may cause the browser to become unresponsive for big previews.
+      if (toVisit.getSize() > 600000) {
+        return distances;
+      }
```
robotsaurus, line 314: What does this mean in terms of the solution?  Will the user still be able to change the tolerance and only get a partial of the flood fill?  It's good to include these ideas in your comments.

charmychancey: Thanks! Will expand the explanation in my in-line comment

PR #66, opened 2020-08-30
scribble with brush size

shmcaffrey:


src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -727,6 +726,42 @@ export class EditorComponent implements OnInit {
...
+    this.floodfillMask(
+      new MaskAction(
+        this.maskTool == MaskTool.MAGIC_WAND_ADD
+          ? Action.ADD
+          : Action.SUBTRACT,
+        Tool.SCRIBBLE,
+        maskSet
+    ));
+  }
+
```
charmychancey, line 764: Hmm, am I correct in understanding that you're trying to merge a lot of what mask-directive handles into editor component? I don't know how merging is gonna work, because I implemented my Preview-Floodfills algo by branching on the current version of master, and that version still uses mask-directive to handle a lot of magicWandService...that means we're gonna have to re-implement my preview-floodfills algo on the UI side, which is gonna require both of us at this point, because you won't be up-to-date with the functions and code that I added, and I may not understand all the implications of the changes that you've made here too. Something to keep in mind for Tues 1|Sept|2020...

src/main/ngapp/accio/src/app/editor/mask.directive.ts
```diff
@@ -173,46 +172,17 @@ export class MaskDirective {
...
-        new Mask.MaskAction(
-          this.tool == MaskTool.MAGIC_WAND_ADD
-            ? Mask.Action.ADD
-            : Mask.Action.SUBTRACT,
-          Mask.Tool.MAGIC_WAND,
-          maskPixels
-        )
-      );
+      this.floodfillEvent.emit(new Coordinate(this.coord[0], this.coord[1]));
     }
```
charmychancey, line 184: #67 Same here

shmcaffrey: Right, probably something we will have to do once yours is branched into master. I knew it would be an issue I was just hoping it wouldn't be too bad. I think we could figure it out depending on how much time there is tomorrow.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -855,7 +887,7 @@ export class EditorComponent implements OnInit {
   
   /** Calls the undo/redo 'do' function with paintedMask: Set of imageData indexes. */
   async doMaskActionPaint(): Promise<void> {
-    let paintedMask = await this.maskControllerPaint();
+    let paintedMask = await this.getPaintedSet();
```
snazzorama, line 890: Is `await` needed now that `getPaintedSet()` isn't `async`?

src/main/ngapp/accio/src/app/editor/mask.directive.ts
```diff
@@ -166,46 +165,18 @@ export class MaskDirective {
...
-        new Mask.MaskAction(
-          this.tool == MaskTool.MAGIC_WAND_ADD
-            ? Mask.Action.ADD
-            : Mask.Action.SUBTRACT,
-          Mask.Tool.SCRIBBLE,
-          maskPixels
-        )
-      );
-    } else if (
+    else if (
```
snazzorama, line 171: nit: `else if` should be on the previous line with the closing brace

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -761,6 +761,42 @@ export class EditorComponent implements OnInit {
...
+      new MaskAction(          
+        this.maskTool == MaskTool.MAGIC_WAND_ADD
+            ? Action.ADD
+            : Action.SUBTRACT,
+          Tool.MAGIC_WAND,
+          maskSet
+    ));
+  }
+
+  async getScribbleSet(coord: Coordinate): Promise<void> {
```
snazzorama, line 781: This function specifies a return type but never returns anything. Is that intentional?

shmcaffrey: the function has to wait for the painted set to be loaded before the rest of the function runs, so it has to be async, and returns void.

src/main/ngapp/accio/src/app/editor/editor.component.ts
shmcaffrey: Chance pointed out that this function, or one that also called getpaintedset() didn't work on his end so I added that it returns a promise.

src/main/ngapp/accio/src/app/editor/editor.component.ts
```diff
@@ -1078,7 +1109,7 @@ class Rectangle {
     this.right = Math.min(coord.x + this.brushRadius, this.imageWidth - 1);
   }
 
-  /** Compares the index's left, right, top, and bottom most pixel based on the brush radius to the current max and mins of all members */
+  /** Compares the indiex's left, right, top, and bottom most pixel based on the brush radius to the current max and mins of all members */
```
snazzorama, line 1112: revert this line ("index" now misspelled)

PR #68, opened 2020-09-01
small bug fixes in master

shmcaffrey:
Fixing three bugs from master that escaped the last two Pr's: 
Issue where the Zoom/pan hot keys weren't working: a 'break' was missing.
Issue where if the image has a mask, the mask isn't drawn when it was initially clicked
Issue where if the image has a mask, and the user presses undo, the initial mask disappears.


charmychancey:
lgtm

PR #70, opened 2020-09-01
Preview to directive refactor

charmychancey:
Change preview-to-frontend implementation to handle directive-refactor:
- Refactor magicWandService call for getPreview() from mask directive into editor component
  - Move into getFloodfillSet()
- Refactor toolbar component to handle toggling between preview floodfill and basic floodfill


dtjanaka:
One issue, I made a separate enum 'UndoRedo', which will need to be addressed when I merge my code into master

